PR #8, opened 2022-08-02
feat: introduce a formatter that prepends "#pragma once"

sxlijin:


trunk.yaml
```diff
@@ -1,6 +1,17 @@
...
   definitions:
+    # Inserts "#pragma once" if it wasn't there
+    - name: pragma-once
+      files: [c-header, c++-header]
+      commands:
+        - output: rewrite
+          run: ${plugin}/linters/pragma-once/fix.sh ${target}
+          formatter: true
+          in_place: true
+          batch: false
```
det, line 12: Can omit batch, its default false.

det (approved):

PR #10, opened 2022-08-05
feat: add cspell linter

TylerJang27:
Add linter definition for [cspell](https://cspell.org/).

Will add PHP, C#, and LaTeX files into definition later after defining in primary trunk

Sample output: 
`/.../run.cc:120:33 - Unknown word (compelete)`


det (approved):
lgtm

PR #20, opened 2022-09-12
chore: trunk.yaml -> plugin.yaml

sxlijin:
Plugin configurations are now shared using plugin.yaml, not trunk.yaml


EliSchleifer (approved):
Nice

PR #25, opened 2022-09-15
Move most trunk actions to plugins repo

EliSchleifer:
Move the trunk specific actions into the plugins repo. 
Because of our merging design these can be defined in both places without any trouble and after we make a new plugins release we can remove the definitions from the built in default_actions file.


det (approved):
lgtm

dapirian (approved):

PR #27, opened 2022-09-17
Disable cspell

TylerJang27:
We have not yet fully audited cspell for use in this particular repo. In order to unblock other PRs, I'm disabling it here.

Future work should include doing a more comprehensive audit to create a configuration/ruleset that would be more appropriate for cspell in this repo.


TylerJang27:
False alarm. We have a specific dictionary for cspell already set up. For the short term, any pain points with cspell should be addressed by updating that dictionary. A long-term solution could include more flexible configuration or logic to auto-add to dictionary.

PR #31, opened 2022-10-04
Modify plugin.yaml to define definitions recursively

EliSchleifer:
Each defined linter/action now has its own plugin.yaml file. This makes actions/linters easy to copy/paste and helps avoid generating a giant long plugin.yaml file which will not scale properly.


EliSchleifer:
@tasgon implemented recursive plugin.yaml support - so the honor is his to review this PR. 

The end result is a much cleaner layout of content

plugin.yaml
```diff
@@ -2,128 +2,16 @@ version: 0.1
...
 actions:
   definitions:
-    - id: commitlint
-      runtime: node
-      packages_file: actions/commitlint/package.json
-      run: commitlint --edit ${1}
-      triggers:
-        - git_hooks: [commit-msg]
-
     - id: git-lfs
```
tasgon, line 5: Is there a reason this and `buf-gen` are still in the plugin root? Since these are both actions, why not put this in actions/ or have a misc folder in there?

EliSchleifer: Good call.

tasgon (approved):
LGTM

PR #32, opened 2022-10-05
chore: Move configs into .trunk/configs 

EliSchleifer:
Take ad


commitlint.config.js
```diff
@@ -0,0 +1 @@
+module.exports = { extends: ["@commitlint/config-conventional"] };
```
TylerJang27, line 1: Were you unable to move this into configs? I'm not as familiar with our node runtimes on this point

TylerJang27 (commented):
LGTM. You may want to change the PR title to `.trunk/config`

TylerJang27 (approved):

EliSchleifer:
I don't think it works to move yet for node side. We have some work to do
to make this work correctly.

PR #34, opened 2022-10-13
feat(actions): add `go-mod-tidy` and `go-mod-tidy-vendor` plugins

sheldonhull:
- `go-mod-tidy` for enabling action on file change for `go.mod`.
- `go-mod-tidy-vendor` for enabling action on file change for `go.mod` that also runs go mod vendor afterward. This helps improve vendoring workflow to consistent and avoid inconsistent vendoring issues by forgetting.

executable bit is set on both scripts.

### DIFF
--- a/actions/go-mod-tidy-vendor/plugin.yaml
+++ b/actions/go-mod-tidy-vendor/plugin.yaml
@@ -4,6 +4,7 @@ actions:
     - id: go-mod-tidy-vendor
       display_name: Go Mod Tidy & Vendor
       description: Runs go mod tidy followed by go mod vendor.
+      runtime: go
       run: ${plugin}/actions/go-mod-tidy-vendor/go-mod-tidy-vendor.sh
       triggers:
         - files: [go.mod]
--- a/actions/go-mod-tidy-vendor/plugin.yaml
+++ b/actions/go-mod-tidy-vendor/plugin.yaml
@@ -1,0 +1,10 @@
+version: 0.1
+actions:
+  definitions:
+    - id: go-mod-tidy-vendor
+      display_name: Go Mod Tidy & Vendor
+      description: Runs go mod tidy followed by go mod vendor.
+      run: ./go-mod-tidy-vendor.sh
+      import_to_global: true
+      triggers:
+        - files: [go.mod]
--- a/actions/go-mod-tidy-vendor/plugin.yaml
+++ b/actions/go-mod-tidy-vendor/plugin.yaml
@@ -4,7 +4,6 @@ actions:
     - id: go-mod-tidy-vendor
       display_name: Go Mod Tidy & Vendor
       description: Runs go mod tidy followed by go mod vendor.
-      run: ./go-mod-tidy-vendor.sh
-      import_to_global: true
+      run: ${plugin}/actions/go-mod-tidy-vendor/go-mod-tidy-vendor.sh
       triggers:
         - files: [go.mod]
--- a/actions/go-mod-tidy/plugin.yaml
+++ b/actions/go-mod-tidy/plugin.yaml
@@ -4,7 +4,7 @@ actions:
     - id: go-mod-tidy
       display_name: Go Mod Tidy
       description: Runs go mod tidy when changes are detected to go.mod.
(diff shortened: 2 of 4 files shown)


actions/go-mod-tidy/plugin.yaml
```diff
@@ -0,0 +1,10 @@
+version: 0.1
+actions:
+  definitions:
+    - id: go-mod-tidy
+      display_name: Go Mod Tidy
+      description: Runs go mod tidy when changes are detected to go.mod.
+      run: ./go-mod-tidy-vendor.sh
```
sxlijin, line 7:
```suggestion
      run: go mod tidy
```

actions/go-mod-tidy-vendor/plugin.yaml
```diff
@@ -0,0 +1,10 @@
+version: 0.1
+actions:
+  definitions:
+    - id: go-mod-tidy-vendor
+      display_name: Go Mod Tidy & Vendor
+      description: Runs go mod tidy followed by go mod vendor.
+      run: ./go-mod-tidy-vendor.sh
+      import_to_global: true
```
sxlijin, line 8: I don't think this field exists on action definitions, but we also don't have good CI checks on PRs into this repo yet.

sxlijin (commented):
What do you think about specifying `runtime: go` on these actions?

IIRC for your use case you use your system runtime for most things.

sheldonhull:
You can't chain go mod vendor after tidy with go runtime. Go mod tidy could be though. Want to me to adjust that one?

actions/go-mod-tidy-vendor/plugin.yaml
```diff
@@ -0,0 +1,10 @@
+version: 0.1
+actions:
+  definitions:
+    - id: go-mod-tidy-vendor
+      display_name: Go Mod Tidy & Vendor
+      description: Runs go mod tidy followed by go mod vendor.
+      run: ./go-mod-tidy-vendor.sh
```
sxlijin, line 7:
```suggestion
      run: ${plugin}/actions/go-mod-tidy-vendor/go-mod-tidy-vendor.sh
```

sxlijin:
> You can't chain go mod vendor after tidy with go runtime. Go mod tidy could be though. Want to me to adjust that one?

Yeah, let's make that change, and then I can go and test this change manually.

(You can do that yourself by specifying this as a local plugin, but there's a bug in how plugin configs work right now. We've fixed it but haven't released the fix yet.)

actions/go-mod-tidy/plugin.yaml
```diff
@@ -4,7 +4,7 @@ actions:
     - id: go-mod-tidy
       display_name: Go Mod Tidy
       description: Runs go mod tidy when changes are detected to go.mod.
-      run: ./go-mod-tidy-vendor.sh
-      import_to_global: true
+      runtime: go
+      run: mod tidy
```
sxlijin, line 8:
```suggestion
      run: go mod tidy
```

dapirian: I think you still need this change for it to work

EliSchleifer: I think the intention here is to call go-mod-tidy.sh 

So probably want the run to be 
go-mod-tidy.sh

EliSchleifer: Or this really
      run: ${plugin}/actions/go-mod-tidy/go-mod-tidy.sh

dapirian (approved):

actions/go-mod-tidy-vendor/plugin.yaml
```diff
@@ -0,0 +1,9 @@
+version: 0.1
+actions:
+  definitions:
+    - id: go-mod-tidy-vendor
+      display_name: Go Mod Tidy & Vendor
+      description: Runs go mod tidy followed by go mod vendor.
```
dapirian, line 6:
```suggestion
      description: Runs go mod tidy followed by go mod vendor.
      runtime: go
```

dapirian (approved):

PR #35, opened 2022-10-14
feat(linters): set up trivy

sxlijin:
Trivy is a comprehensive security scanner: https://github.com/aquasecurity/trivy


laurit17 (approved):
feel like cspell mostly gets in the way...

PR #37, opened 2022-10-17
feat(linters): implement sqlfmt

sxlijin:
This is a black-style formatter that is an alternative to sql-formatter for people.


laurit17 (approved):

tconbeer:
Hi @sxlijin, I'm the creator of sqlfmt. Thanks so much for creating this integration!

I don't know anything about Trunk, but I have a couple of suggestions:

1. sqlfmt works best when installed with the jinjafmt extra, e.g., `pip install shandy-sqlfmt[jinjafmt]`. Is it possible to use this extra from Trunk? Would `packages: shandy-sqlfmt[jinjafmt]` do the trick?
2. sqlfmt logs most of its output to `stderr`. We only use `stdout` if formatting code passed through `stdin`, e.g., `echo "select 1" > sqlfmt -`. So I think `read_output_from: stderr` would be better?

Finally, I'd love to mention this integration in our [documenation](https://docs.sqlfmt.com/category/integrations). Is this how someone would enable sqlfmt?

```yml
# in trunk.yaml
lint:
  enabled:
    - sqlfmt@0.13.0
```

tconbeer:
PR is here: https://github.com/trunk-io/plugins/pull/45

Thanks again for your help with this!

PR #38, opened 2022-10-17
fix(actions): disable commitlint

sxlijin:
commitlint is consistently broken for me and just screws up my commits every single time i make one

plus it doesn't make sense as a commit hook for anyone using squash-and-merge workflows: in that scenario the thing that actually matters is the merge description.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -47,7 +47,6 @@ actions:
             - lint/**
           run: tools/toggle-local/toggle notify ${workspace}
   enabled:
-    - commitlint
     - toggle-local
     - trunk-cache-prune
     - trunk-upgrade-available


EliSchleifer:
What is the commitlint rule equivalent for a pull request?

EliSchleifer (approved):

sxlijin:
That's where the MQ comes in, I think... I don't know if there's a way around it.

Also why I filed https://linear.app/trunk/issue/TRUNK-4681/set-up-a-pre-commit-hook-in-trunk-ioplugins

EliSchleifer:
Maybe when trunk push creates a pull request - it can recommend / push
through a flow to do things in a conventional format. That's what I want

PR #39, opened 2022-10-18
feat(linters): set up nancy, a go security linter

sxlijin:
Nancy checks the Go dependency graph against the Sonatype OSS index.

Nancy comes in two flavors, sleuth and IQ, but only sleuth advertises a JSON output format, so we leave IQ alone for now.

Also, sleuth can run into rate limits that require authz tokens to bypass; see
https://github.com/sonatype-nexus-community/nancy#oss-index-options.

### DIFF
--- a/linters/nancy/parse.py
+++ b/linters/nancy/parse.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+
+"""
+nancy sleuth --output=json looks like this:
+
+```
+{
+  ...,
+  "vulnerable": [
+    {
+      "Coordinates": "pkg:golang/golang.org/x/text@v0.3.7",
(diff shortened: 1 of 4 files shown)


linters/nancy/parse.py
```diff
@@ -0,0 +1,88 @@
...
+  ]
+}
+```
+"""
+
+import json
+import sys
+
+
+def to_result_sarif(
```
laurit17, line 36: general question - do we have/plan to have any kind of type safety here? I imagine we could define a sarif type and then annotate this to match it quite easily.

sxlijin: we should, but i'm not familiar with mypy or how to set up pytype

laurit17 (approved):

PR #42, opened 2022-10-21
feat: Add action to sort cspell-words dictionary on save

EliSchleifer:
Tiny little action to save a sorted version of cspell-words.txt on save


EliSchleifer:
Weird behavior here...on push...we are seeing the pre-push trunk check run detect a discarded file.
image

dapirian (commented):
Would this be better as a formatter?

EliSchleifer:
Yes I rewrote this as a formatter - kept this PR because it exposes a bug

dapirian (approved):

PR #52, opened 2022-12-01
fix(codespell): missed direct_configs and affects_cache

sxlijin:
(have not yet tested this)


laurit17 (approved):

sxlijin:
Manually tested, confirmed that `.codespellrc` is now propagated into the sandbox.

PR #53, opened 2022-12-01
consolidate sqlfluff definitions

laurit17:
It looks like this bug has been fixed since it was filed; we can replace the definition - https://linear.app/trunk/issue/TRUNK-4669/sqlfluff-fix-how-mutliple-commands-are-resolved-for-formatters-and


EliSchleifer (approved):
LGTM

PR #57, opened 2022-12-05
fix(sqlfluff): success_codes or error_codes must be set

sxlijin:
Tested already.


sxlijin:
Sent this before I saw #56

PR #65, opened 2022-12-15
Added CircleCI configuration linter

docwhat:
This is a simple linter definition for the `circleci` CLI tool.

I removed the `<!-- trunk-ignore(markdownlint/MD041) -->` because trunk told me it wasn't needed.  If that was wrong, I could put it back.

I'm really liking the `trunk` tool and appreciate how well-designed the tool is.

I would suggest adding a `CONTRIBUTING.md` file with instructions for what you would expect for a well-formed linter or action.

PR #68, opened 2022-12-15
Improve testing workflow

TylerJang27:
Changes targeted by this PR:
- Avoid snapshot assertions failing when it's a new version that doesn't have a snapshot. This includes removing the requirement to specify output files, and now snapshots are always autogenerated. See comment in `linterCheckTest()` for specifics on this design. This decision was rooted in a couple factors, namely:
  - As identified in #65, the lack of a snapshot for a specific version was causing unnecessary failures.
  - Linters can still change their messages, outputs, etc (as sqlfluff did yesterday). This is something that we don't want to cause too much friction, but we still want to call it out as a noticeable change. If this happens during someone's workflow (should still be relatively infrequent), now simply run the specific command in tests/readme.md and validate expected output. _We may want to highlight this specific command better or with more visibility._
  - The JSON assertions and snapshot code were a bit redundant and unnecessarily complex. This should simplify and be flexible enough for the near future. If we want any additional assertions, we can do so via the `postCheck` hook.
- Make tests run in parallel. This also simplifies the setup for the `config_check_test.ts`. It's a tad redundant, but its presence makes triaging problems a bit easier.
- Attempt to reduce flakiness (the parallelism appears to reduce flakiness to about 1-3%--it'd be nice to have analytics on this). I'll continue investigating alternatives for dealing with the daemon lifetime.

Average test time right now locally with these 4 tests is about 10 seconds in total :)
They're about a minute in total on CI because of the need to download and install linters.

Remaining todos (unordered):
- Investigate daemon lifetime changes
- Modify sandboxing to shadow tree from root
- Use debugging library in testing
- Add contributing.md
- Add option for specifying a range of linter versions (ideally for nightlies)
- Migrate rest of definitions


.trunk/trunk.yaml
```diff
@@ -36,7 +36,7 @@ lint:
     - linters:
         - ALL
       paths:
-        - linters/**/test/*
+        - linters/**/test*/*
```
TylerJang27, line 39: Allows for more lenience (e.g. `test` vs. `tests`)

linters/pragma-once/test/__snapshots__/pragma-once_basic_fmt.shot
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
TylerJang27, line 1: If desired, I can move the filepath to not nest this inside of __snapshots__ so that it's more apparent, although I think having autogenerated files in a subdirectory (as is right now) is probably the better move.

tests/utils/landing_state.ts
```diff
@@ -1,82 +1,41 @@
...
-  paths,
-  linter,
-  parser,
-  report,
-  upstream,
-  fileGroupName,
-  command,
-  verb,
-  ..._rest
+  actionDurationMs: _actionDurationMs,
```
TylerJang27, line 7: Ideally this would just be a matcher rather than just removing it, but I was having a bit of difficulty having the matcher play nice with the `jest-specific-snapshot` dependency I was using. Will revisit if needed, but this should be fine for now.

linters/pragma-once/test/__snapshots__/pragma-once_basic_fmt.shot
sxlijin: agree w you

tests/utils/landing_state.ts
```diff
@@ -1,82 +1,41 @@
...
+}: LandingState) =>
   <LandingState>{
-    issues: sort(issues.map((i) => extractFileIssueFields(i, skip_message))).asc([
+    issues: sort(issues).asc([
       (issue) => issue.file,
       (issue) => issue.line,
       (issue) => issue.column,
       (issue) => issue.code,
       (issue) => issue.message,
     ]),
```
sxlijin, line 37:
```suggestion
    issues: sort(issues).asc(issue => [issue.file, issue.line, issue.column, issue.code, issue.message]),
```

sxlijin (commented):
need to do another pass tmrw to understand snapshots better

tests/index.ts
```diff
@@ -116,36 +102,27 @@ export const linterCheckTest = ({
     const driver = setupDriver(dirname, {}, linterName, preCheck);
 
     // Step 3: Asynchronously run each test
```
sxlijin, line 104: FYI this does not actually run tests asynchronously: it simply tells mocha that there's async logic in the test case and that it needs to `await` the result of each test case.

[This SO answer](https://stackoverflow.com/a/35534550) goes into more detail, but tldr is that mocha is sequential by design (and you can guarantee which test case an exception was thrown inside). Glancing at mocha-parallel-tests it may help, but not sure. We may need to write our own version of mocha-parallel-tests.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "trunk": "trunk",
-    "test": "jest --runInBand"
+    "test": "jest"
```
sxlijin, line 6: `npm test` does not work for me; `jest` is not installed into my `PATH` when I `npm install` locally.

```
sam@sam-dell:2022-12-19T18:20:22-0800:~/plugins2 (tyler/improve-testing-1) [1]
$ npm test
++ 2022-12-19T18:20:24-0800 ++ npm test

> trunkio-plugins@1.0.0 test
> jest

sh: 1: jest: not found
```

(maybe there's something wrong with my personal `PATH` setup and npm/PS1 hooks? not sure)

tests/utils/index.ts
```diff
@@ -1,4 +1,6 @@
+import fs from "fs";
```
sxlijin, line 1: An FYI comment re `utils` as this dir name: https://go.dev/blog/package-names#bad-package-names (I think this structure is fine as is right now).

(Personally, I think a kitchen-drawer library is always necessary, since it removes some amount of mental decision making around "where should this go" for stuff that has no clear home; but I do agree that we should always be looking for a path to move things out of the kitchen sink.)

tests/readme.md
```diff
@@ -69,6 +64,18 @@ To run an individual test, run:
 npm run test ${path_to_linter_subdir}
```
sxlijin, line 64:
```suggestion
npm test ${path_to_linter_subdir}
```

tests/readme.md
```diff
@@ -86,3 +93,6 @@ Options include:
 - `PLUGINS_TEST_CLI_PATH` specifies an alternative path to a trunk binary
 - `PLUGINS_TEST_LINTER_VERSION` specifies a linter version semantic (KnownGoodVersion | Latest |
   version)
+- `PLUGINS_TEST_NEW_SNAPSHOT` if "true" tells tests to use an exact match of the linter version when
```
sxlijin, line 96:
```suggestion
- `PLUGINS_TEST_UPDATE_SNAPSHOTS` if "true" tells tests to use an exact match of the linter version when
```

linters/todo-linter/test/__snapshots__/todo-linter_basic_check.shot
```diff
@@ -0,0 +1,51 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
sxlijin, line 1: Is there any way we can change the format of this prefix? Like:

> This file is autogenerated by Jest. See tests/readme.md#update for more details.

tests/utils/index.ts
```diff
@@ -26,7 +28,64 @@ export const ARGS: TestingArguments = {
   cliVersion: process.env.PLUGINS_TEST_CLI_VERSION,
   cliPath: process.env.PLUGINS_TEST_CLI_PATH,
   linterVersion: parseLinterVersion(process.env.PLUGINS_TEST_LINTER_VERSION ?? ""),
+  dumpNewSnapshot: process.env.PLUGINS_TEST_NEW_SNAPSHOT?.toLowerCase() === "true",
```
sxlijin, line 31:
```suggestion
  dumpNewSnapshot: process.env.PLUGINS_TEST_NEW_SNAPSHOT?.toLowerCase() === "true",
```

just test if the envvar is set to something that isn't false-y - setting `ENV=1` is a common way to activate a given envvar

tests/readme.md
```diff
@@ -86,3 +93,6 @@ Options include:
 - `PLUGINS_TEST_CLI_PATH` specifies an alternative path to a trunk binary
 - `PLUGINS_TEST_LINTER_VERSION` specifies a linter version semantic (KnownGoodVersion | Latest |
```
sxlijin, line 94: Now that I actually read this... this is a bit confusing, because this propagates to every single linter test running. In other words, if I do

```
PLUGINS_TEST_LINTER_VERSION=1.4.4 npm test
```

It applies 1.4.4 to every linter we're testing, and not just sqlfluff.

I think if we use a schema where we use snapshot names to define the test cases, that will solve this problem.

tests/utils/index.ts
```diff
@@ -26,7 +28,64 @@ export const ARGS: TestingArguments = {
...
+  }
+
+  // If this is a versioned linter && dumpNewSnapshot, return its generated name.
+  const specificVersionSnapshotName = path.resolve(
+    snapshotDirPath,
+    `${linterName}_v${linterVersion}_${prefix}.shot`
+  );
+  if (ARGS.dumpNewSnapshot) {
+    return specificVersionSnapshotName;
+  }
```
sxlijin, line 64: I think that if we want to update the snapshot in question, we can rename `linter_vLatest.out` to `linter_vX.Y.Z.out` and overwrite `linter_vLatest.out`. It seems that right now if the latest version of the linter changes on you, you need to do some manual work _after_ you re-run with `PLUGINS_TEST_NEW_SNAPSHOT=true`?

tests/config_check_test.ts
```diff
@@ -10,24 +10,12 @@ const subfolder = path.resolve(REPO_ROOT, "linters");
...
     setupGit: false,
-    setupTrunk: false,
-    launchDaemon: true,
+    setupTrunk: true,
+    launchDaemon: false,
   });
 
   // Step 2: Validate config
   it("trunk config print from repo root", async () => {
-    // trunk-ignore(eslint/@typescript-eslint/no-unsafe-assignment)
```
sxlijin, line 19: Is the reason we're doing this because we want to do the driver matrix testing at the workflow level?

sxlijin (commented):
So this is a bit of a long one- I admit that I didn't review the previous ones very carefully, and was doing more of a sanity-check pass over them. Which is why this one is so long 😅 

Overall I like the direction this is going in: I like that we're re-using jest snapshots for the purpose of this, since it looks like a good fit here even if we're not doing UI testing.

The two big points:

1. I'm a bit confused about the workflow for testing old linter versions / updating when a linter releases a new version with new output, and
2. I think we can simplify the test data dir structure.

.trunk/trunk.yaml
TylerJang27: __snapshots__ is autogenerated in the original package, but when specifying a path in `jest-specific-snapshot`, you have to specify a destination, so I stuck to the convention.

tests/index.ts
TylerJang27: The comment itself is out of date 🙃 . I'll update. Based on some discussion with Dylan, I don't think it's natively possible/feasible in jest to run each it() in parallel, just use parallelism at the top level (per test file). I think the slowdown isn't too significant, as long as you don't have a large number of tests in one test file, so I think it's fine to keep as is.

package.json
TylerJang27: Let's figure this out

tests/utils/index.ts
TylerJang27: Had the same train of thought. I think right now it's fine, but I intend to extract things out as it expands and begins to take on the flavor of any particular category.

tests/config_check_test.ts
```diff
@@ -10,24 +10,12 @@ const subfolder = path.resolve(REPO_ROOT, "linters");
 describe("Global config health check", () => {
   const driver = setupDriver(subfolder, {
     setupGit: false,
-    setupTrunk: false,
-    launchDaemon: true,
+    setupTrunk: true,
+    launchDaemon: false,
```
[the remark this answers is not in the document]
TylerJang27: Kinda. Has to do with versioning, and I changed what the test does. Before, it had the potential to be a different version of the daemon than the other tests, which would have caused problems when running in parallel. This new setup guarantees they all have the same daemon issue, which allows for parallelism, but doesn't fix the timeout issue.

tests/readme.md
TylerJang27: Not sure I follow. Let's discuss

tests/utils/index.ts
TylerJang27: I think there's a slight gap in understanding here. Let's discuss

tests/readme.md
TylerJang27: For future: If PLUGINS_TEST_LINTER_VERSION is version, add todo for make sure only run with one test

TylerJang27: Also todo: callout -u/our behavior for the flag

sxlijin (approved):

tests/utils/index.ts
TylerJang27: TODO: Use same pattern with formatters and linters

TylerJang27: TODO: Callout snapshot resolution in readme.

TylerJang27:
Addressing all comments in a follow-up to unblock folks with testing failures

tests/index.ts
```diff
@@ -184,21 +161,21 @@ export const linterFmtTest = ({
...
-          }),
         });
 
-        const expectedOutFile = path.resolve(dirname, path.parse(outputPath).base);
+        const formatSnapshotPath = path.resolve(
+          dirname,
+          SNAPSHOT_DIR,
+          `${linterName}_${prefix}_fmt.shot`
+        );
         // trunk-ignore(eslint/@typescript-eslint/no-non-null-assertion)
```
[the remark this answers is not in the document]
TylerJang27: I think having this rule is good, I just want to override it in this case because its existence is part of the assertion.

PR #69, opened 2022-12-20
Improve testing workflow further, add debug and coverage

TylerJang27:
This is a followup to #68 and introduces the following:
1. Added debug library (instructions for use in `tests/readme.md`). Namespaces are of form: `${Location}/${linter}/${#}`.
2. Add test that asserts linter test coverage is required.
3. Add `Snapshots` option to `PLUGINS_TEST_LINTER_VERSION` that runs tests on each snapshot version that is provided (eventually we can run this on nightlies, but for now it's a useful tool for testing new cli versions).

Also addressed are:
1. Linters and formatters now have unified behavior when it comes to checking snapshot versions. They may not necessarily provide the same version set, however, since formatters tend to be more stable.
2. Modified test directory structure to use a `test_data` folder and no `__snapshots__` folder. See the modifications in `tests/readme.md`.
3. Added information about snapshot resolution to `tests/readme.md`.
4. Rename `PLUGINS_TEST_NEW_SNAPSHOT` to `PLUGINS_TEST_UPDATE_SNAPSHOTS`.
5. Rename snapshot files to syntax: `${linter}_v${linterVersion}_${prefix}.${check/fmt}.shot`.
6. Added comments for clarity throughout.


tests/test_coverage_test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { REPO_ROOT } from "tests/utils";
+
+// TODO(Tyler): Burndown this list (currently all linters defined before the testing framework was implemented).
+const excludedLinters: string[] = ["codespell", "cspell", "nancy", "oxipng", "sqlfmt", "trivy"];
+
+// This test asserts that all linters have at least one test. All new linters are expected to have
+// test coverage. Review tests/readme.md for testing guidelines. Prefer using npm test for indirection
+// in this test so that we get an accurate list of all tests, regardless of any changes to the test spec
+// in jest.config.ts.
+describe("All linters must have tests", () => {
```
sxlijin, line 13: nice!

Spent a little time stewing on the dir structure for `tests/` and I think I'm fine with the current state.

* Generally I prefer for tests to be separate from testing utils, but this also results in `tests/` and `testing/` (which has both a tab completion problem and also a "what's the difference" question for newcomers)
* Most testlib/testutil logic is in index.ts or subdirs, so it's not too much of a difference.
* I wasn't super happy with other names that I was coming up with:
  * `self_test/coverage_test.ts`
  * `trunk/coverage_test.ts`
  * `plugins_test/coverage_test.ts`
  * `meta/coverage_test.ts`
  * (obviously you can mix and match `foo/` and `foo_test/` above)
* The only non-test file we have that lives at the root of `tests/` is `index.ts`, which is... fine, I think.

tests/utils/index.ts
```diff
@@ -23,48 +29,93 @@ const parseLinterVersion = (value: string): LinterVersion | undefined => {
...
+};
+
+/**
+ * Calculate the regex for a given snapshot file to determine available versions. Use this as a standardized convention.
+ * @param linterName the name of the linter being tested. Does not include subcommand
+ * @param prefix the prefix of the named file tested against
+ * @param checkType "check" or "fmt"
+ */
+export const getSnapshotRegex = (linterName: string, prefix: string, checkType: "check" | "fmt") =>
+  `${linterName.replace("-", "_")}_v(?<version>(\\d.?)+)_${prefix}.${checkType}.shot`;
```
sxlijin, line 82:
```suggestion
  `${linterName.replace("-", "_")}_v(?<version>([^_]?)+)_${prefix}.${checkType}.shot`;
```

to match things like 2.12.1-beta

tests/utils/index.ts
```diff
@@ -86,6 +137,33 @@ export const getSnapshotPath = (
...
+        return fileMatch.groups?.version;
+      }
+    })
+    .filter(Boolean);
+
+  if (versionsList.length === 0) {
+    return [undefined];
+  }
+
+  return new Set(versionsList);
```
sxlijin, line 168: can we just return `versionsList` so that we always test in the same order?

tests/readme.md
```diff
@@ -55,44 +54,108 @@ examples.
...
 
 ### Linter Versioning
 
 Missing snapshots will be automatically created based on test input files. If an existing test fails
-because a new linter version has introduced a breaking change, rather than running
-`npm run test -- -u`, **instead run**
-`PLUGINS_TEST_NEW_SNAPSHOT=true npm run test ${path_to_failing_test}`. This is used to track
-historical test behavior and ensure compatibility with trunk across multiple linter versions. See
-"Environment Overrides" below.
+because a new linter version has introduced a breaking change, rather than running `npm test -- -u`,
```
sxlijin, line 72:
```suggestion
because a new linter version has introduced a breaking change, rather than running `npm test -- -u`,
```

Can we detect `-u` and fail?

If not, NBD.

tests/index.ts
```diff
@@ -153,32 +173,44 @@ export const linterFmtTest = ({
...
-    // Step 3: Asynchronously run each test
     linterTestTargets.forEach(({ prefix, inputPath }) => {
-      it(prefix, async () => {
-        const testRunResult = await driver.runFmtUnit(inputPath, linterName);
-        expect(testRunResult).toMatchObject({
-          success: true,
+      // Step 1b: Detect versions to test against if PLUGINS_TEST_LINTER_VERSION=Snapshots
+      const linterVersions = getVersionsForTest(dirname, linterName, prefix, "fmt");
+      linterVersions.forEach((linterVersion) => {
+        // TODO(Tyler): Find a reliable way to replace the name "test" that doesn't violate snapshot export names.
```
sxlijin, line 184: Ah, I take it you did want to put the version in the describe name and then ran into this?

sxlijin (approved):

tests/utils/index.ts
```diff
@@ -23,48 +29,93 @@ const parseLinterVersion = (value: string): LinterVersion | undefined => {
...
+ * @param checkType "check" or "fmt"
  * @param linterVersion version of the linter that was enabled (may be undefined)
  * @returns absolute path to the relevant snapshot file
  */
-export const getSnapshotPath = (
+export const getSnapshotPathForAssert = (
   snapshotDirPath: string,
   linterName: string,
   prefix: string,
+  checkType: "check" | "fmt",
```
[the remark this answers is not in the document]
TylerJang27: Considered a `CheckType = "check" | "fmt"` but semi-arbitrarily decided against it. If/when I revisit that I will add a type for it

tests/test_coverage_test.ts
TylerJang27: Thanks for the thinking here. I suspect we may revisit this a bit in the intermediate future. As for the last bullet point, I liked having the `index.ts` in `tests/`, since it makes imports in the individual linter tests easy--you can just import `linterCheckTest` from `tests` rather than specifying a longer and repetitive path.

tests/utils/index.ts
TylerJang27: There's some de-duping that is happening here in the case of multiple prefixes. I'll add some logic for this for deterministic ordering, and also a comment explaining why

tests/readme.md
TylerJang27: We can... but -u should also be an option, in case we need to update an old snapshot (should be rare, but can and will happen. This is also useful for debugging if you've configured something poorly; e.g. your test runs check, and it generates a taskFailure that you did not intend (right now we don't error out on the presence of taskFailures, in the future we may want to add more test helper functions than just `linterCheckTest` that assert these kinds of failures, such as on empty files).

Overall, we may want to revisit some of the specifics related to this particular workflow in the future, particularly as we see any pitfalls that users are hitting when contributing.

tests/index.ts
TylerJang27: Yep. It gets written to the snapshot, which is deterministic when you're running with `PLUGINS_TEST_LINTER_VERSION=Snapshots`, but not otherwise, since we don't know the version that's going to be enabled yet, causing it to be undefined--and quite annoying.

tests/utils/index.ts
TylerJang27: Nevermind, I restructured in order to pass in a prefix. I'll just make this a list

PR #72, opened 2022-12-23
Cache downloads on PR runs

TylerJang27:
Accomplishes two things:
1. Cache linter downloads across PR runs
2. Fixes a small bug with how preCheck and postCheck callbacks are invoked (this is an important fix for getting git-diff-check working).

When running 20+ tests, this brought the PR time from 10+ minutes down to 4 minutes. I still have some timeouts to investigate on large numbers of tests in CI.

I don't think there's a real need to flush/reset these downloads, but when we add nightly workflows we can implement that.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -26,7 +26,7 @@ jobs:
   plugins_test:
     name: Plugins Test Runner
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 20
     strategy:
       matrix:
         linter-version: [KnownGoodVersion, Latest]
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -26,14 +26,21 @@ jobs:
...
       matrix:
         linter-version: [KnownGoodVersion, Latest]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Cache tool downloads
+        uses: actions/cache@v3
+        with:
+          path: /tmp/plugins_testing_download_cache
```
sxlijin, line 40: Where does this get wired into trunk? Inside trunk-io/action?

sxlijin (approved):

.github/workflows/pr.yaml
TylerJang27: It's built-in in driver.ts. Sets `TRUNK_DOWNLOAD_CACHE` to `plugins_testing_download_cache`. The trunk-io/action has its own caching that is auto-handled that is in a different location. We don't have any nightlies yet here, so the clearing logic you mentioned doesn't apply here.

This should probably be wired in more clearly or from a higher level. No immediate ideas, but we can address in a follow-up

PR #74, opened 2022-12-29
Remove unintended debug test code

dapirian:


.vscode/extensions.json
```diff
@@ -1,3 +1,3 @@
 {
-  "recommendations": ["eamodio.gitlens", "redhat.vscode-yaml", "trunk.io"]
+  "recommendations": ["redhat.vscode-yaml", "trunk.io"]
```
dapirian, line 2: Just sticking to what literally helps using trunk (the former gives us autocomplete in trunk.yaml)

package.json
```diff
@@ -33,10 +33,10 @@
     "typescript": "^4.9.4",
     "yaml": "^2.2.0"
   },
+  "bundleDependencies": [
+    "tests"
+  ],
```
dapirian, line 38: They accept both `bundle` and `bundled`, but the former is recommended. Either way works

TylerJang27 (approved):

PR #76, opened 2022-12-29
Add sort-package-json

dapirian:
- Tested locally
- We need to upgrade the testing code to support "in" files for formatting that have a specific name, not a prefix (package.json must be named package.json for the tool to accept it)
- This tool has no --version command, or any cmd line args really 😆 

I think I'd like to mark this `good_without_config`, but need to make sure _most_ people will actually want their package.json sorted in this way first.

### DIFF
--- a/linters/sort-package-json/plugin.yaml
+++ b/linters/sort-package-json/plugin.yaml
@@ -1,0 +1,20 @@
+version: 0.1
+lint:
+  files:
+    - name: package-json
+      filenames: [package.json]
+
+  definitions:
+    - name: sort-package-json
+      files: [package-json]
+      commands:
+        - name: format
+          output: rewrite
+          run: sort-package-json ${target}
+          success_codes: [0]
+          in_place: true
+          formatter: true
+      runtime: node
+      package: sort-package-json
+      # good_without_config: true # decide whether this is something we want on by default
+      known_good_version: 2.1.0
--- a/tests/test_coverage.test.ts
+++ b/tests/test_coverage.test.ts
@@ -10,6 +10,7 @@ const excludedLinters: string[] = [
   "cspell",
   "nancy",
   "oxipng",
+  "sort-package-json",
   "sqlfmt",
   "trivy",
 ];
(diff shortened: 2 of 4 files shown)


linters/sort-package-json/plugin.yaml
```diff
@@ -0,0 +1,20 @@
+version: 0.1
+lint:
+  files:
+    - name: package-json
+      filenames: [package.json]
```
TylerJang27, line 5: Defining files in plugins has other side effects, primarily since we don't support namespacing file types (so any plugins that define file types must therefore be globally namespaced). This isn't really a problem, and we already do this already, but we've been defining all file types at the top level plugin.yaml just for visibility and ease of access for the future. Can you move this there?

TylerJang27: Alternatively, there is some benefit to having some proximity to it's main use case (as in the case of the circle ci linter, which does the same thing. Ultimately your call, but wanted to highlight the choice that we had been making up to that point.

tests/test_coverage.test.ts
```diff
@@ -10,6 +10,7 @@ const excludedLinters: string[] = [
   "cspell",
   "nancy",
   "oxipng",
+  "sort-package-json",
```
TylerJang27, line 13: re: your comment in the description about testing. I'm planning to add some better helper methods for more manual trunk check runs (this is going to be particularly necessary for triggers, etc.). I will loop back to this linter then and add some tests for it.

TylerJang27: This will also require some slight modifications to the snapshot format (or at least a new naming schema for those types of tests).

TylerJang27 (approved):
Approving, note my first comment about optional filetype relocation

TylerJang27:
I'm currently working on chasing down the flaky timeout that is sometimes happening in the CI tests. Don't let it block you

linters/sort-package-json/plugin.yaml
dapirian: Leaving it here for now for the same reason as circleci, but if we ever have a 2nd user we should definitely move it to the root

PR #80, opened 2023-01-08
Add nightly job

TylerJang27:
This will run against all existing snapshots AND against the latest version of each linter.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -3,35 +3,13 @@ on:
   schedule:
     # Times in UTC (PST+8), this translates to running Mon-Fri at midnight/1am (depending on DST)
     - cron: 0 8 * * 1-5
-  pull_request: {}
+  pull_request:
+    paths: [.github/workflows/nightly.yaml]
   workflow_dispatch: {}
 
 jobs:
-  check_nightly:
-    name: Should Run Nightly
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    outputs:
-      run-nightly: ${{ steps.changes.outputs.nightly }}
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-
-      - name: Filter changed nightly.yaml
-        uses: dorny/paths-filter@v2
-        id: changes
-        with:
-          filters: |
-            nightly:
-              - '.github/workflows/nightly.yaml'
-
   plugins_test:
     name: Nightly Plugin Tests
-    needs: check_nightly
-    # Only run if triggered from PR and this file was modified OR triggered from workflow_dispatch or nightly
-    if:
-      ${{ needs.check_nightly.outputs.run-nightly }} == 'true' || $GITHUB_EVENT_NAME !=
-      "pull_request"
     runs-on: ubuntu-latest
     timeout-minutes: 60
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,39 @@
...
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        linter-version: [Snapshots, Latest]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+        # TODO(Tyler): Determine if we want to wipe the download cache and start fresh each night.
```
TylerJang27, line 20: Do we want to wipe out the contents of the cache? Is the easiest way to do that just rm?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,39 @@
+name: Nightly
+on:
+  schedule:
```
sxlijin, line 3: nice-to-have: also run this workflow on changes to this file (we use paths-filter for this in the monorepo's `pr.yaml`)

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run plugin tests
+        run: |
+          PATH=$(npm bin):$PATH PLUGINS_TEST_LINTER_VERSION=${{ matrix.linter-version }} npm test --ci
```
sxlijin, line 39: i think using the native mechanism where possible is cleaner here:

```suggestion
      - name: Run plugin tests
        run: |
          PATH=$(npm bin):$PATH npm test --ci
        env:
          PLUGINS_TEST_LINTER_VERSION: ${{ matrix.linter-version }}
```

.github/workflows/nightly.yaml
sxlijin: Pretty much, yeah. Look at .github/actions/linux_build_and_test/action.yaml in the monorepo for an exaple.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+        # TODO(Tyler): Determine if we want to wipe the download cache and start fresh each night.
+
+      - name: Cache tool downloads
+        uses: actions/cache@v3
+        with:
+          path: /tmp/plugins_testing_download_cache
+          # No need to key on trunk version unless we change how we store downloads.
```
sxlijin, line 26:
```suggestion
          # No need to key on trunk version or trunk.yaml unless we change how we store downloads.
```

in comparison to https://github.com/trunk-io/trunk-action/blob/2cda8a0c9923423307c2c3efe676a4c4614e46e6/action.yaml#L88-L91

sxlijin (approved):

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,76 @@
+name: Nightly
+on:
+  schedule:
+    # Times in UTC (PST+8), this translates to running Mon-Fri at midnight/1am (depending on DST)
+    - cron: 0 8 * * 1-5
+  pull_request: {}
```
sxlijin, line 6:
```suggestion
  pull_request:
    paths: [.github/workflows/nightly.yaml]
```

looked it up after, looks like you can apply the filter on the trigger itself

PR #87, opened 2023-01-12
Add mac-specific linters and skip tests conditionally

TylerJang27:
5 major changes:
1. Run CI on both Mac and Ubuntu
2. Adds optional skipTestIf parameter to custom test interfaces. Used for skipping tests on certain OS or certain combinations of OS and linter version (imperfect, but should suffice for snapshot runs)
3. Add swiftlint, stringslint, swiftformat
4. Disable ruby runtime tests on Mac
5. Bump test version (as determined by repo root trunk.yaml) to include the timeout bug fix.

Prefer skipping a test over an early return or success so that it's highlighted for visibility. There weren't any built-in functions for skipping a test when inside its scope that I could find, but this appears to be a suitable alternative to conditionally call it.skip.

See skip output in [CI job](https://github.com/trunk-io/plugins/actions/runs/3898270796/jobs/6656834209)


.github/workflows/nightly.yaml
```diff
@@ -10,12 +10,15 @@ on:
...
   plugins_test:
     name: Nightly Plugin Tests
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, macos-latest]
         linter-version: [Snapshots, Latest]
```
TylerJang27, line 19: I was considering adding some config to prevent some Nightly jobs from running when the trigger is `pull_request` and the linter-version is `Latest`, since there is an identical run happening from `pr.yaml`, but not sure if that is too complex and/or straying from "repeatable runs" with those of nightly

linters/brakeman/brakeman.test.ts
```diff
@@ -1,4 +1,7 @@
 import { customLinterCheckTest } from "tests";
-import { TEST_DATA } from "tests/utils";
+import { osTimeoutMultiplier, skipOS, TEST_DATA } from "tests/utils";
```
det, line 2: `skipOS` is wrong

det (approved):

det (approved):

PR #90, opened 2023-01-17
Migrate detekt-gradle

TylerJang27:
Migrate detekt-gradle. Tests are limited due to dependence on the jar file, but this should at least increase visibility and be the last one we need to port.


linters/detekt/detekt.test.ts
```diff
@@ -11,3 +14,17 @@ linterCheckTest({ linterName: "detekt", namedTestPrefixes: ["basic_detekt"], pre
 
 // detekt-explicit has no default settings, leading to an empty result
 linterCheckTest({ linterName: "detekt-explicit", namedTestPrefixes: ["basic_explicit"], preCheck });
+
+// detekt-gradle tests rely on a particular gradle-focused system setup, and they require a level of greater configuration here.
+// TODO(Tyler): Because of this setup, this leads to reduced coverage for different versions. We should augment this logic
+// without using a static jar.
+const gradlePreCheck: TestCallback = (driver) => {
+  // Move files to root
```
sxlijin, line 22: add some more comments about the requirements: namely, that plugin.yaml and gradlew both need to be available. discuss alt. impl. options (e.g. chdir to the test_data) and explain why this was chosen.

sxlijin (approved):

PR #96, opened 2023-01-22
Improve TaskFailure debug experience

TylerJang27:
The failure `detailPath` was formerly omitted, but passing in asymmetric wrappers to `toMatchSpecificSnapshot` provides an alternative that prints much more detailed failure information. 

### Background
Improves CI and local experience of debugging when failures occur. This has been a long time coming. Previous [failure runs](https://github.com/trunk-io/plugins/actions/runs/3977200383/jobs/6818150578) were painfully difficult to debug, particularly if they couldn't be reproduced locally, and would often involve `cd`-ing into autodeleted directories. Now, failure detail information is printed directly through the use of asymmetric property matchers (see screenshot below).

Screenshot 2023-01-21 at 11 39 16 PM

### Alternatives (that I considered and attempted)
1. Using custom serializers. [jest-specific-snapshot](https://github.com/igor-dv/jest-specific-snapshot) supports this, and there are decent [documentation](https://jestjs.io/docs/expect#expectaddsnapshotserializerserializer) and [guides](https://medium.com/@luisvieira_gmr/jest-snapshot-serializers-6a96f5c362a1) about how to approach this. However, whatever serialized output is created is written to the snapshot file itself. This means that we would have to save all of the detailPath contents (which presents numerous problems, including nondeterministic paths), or none at all.
2. Ditching snapshots completely and using a stricter JSON approach. This would be closer to the original approach, but it flies in the face of recent work, including numerous QoL improvements, so I did not thoroughly consider this option.
3. Using a custom asymmetric property matcher. There is also decent documentation on [property matchers](https://jestjs.io/docs/snapshot-testing#property-matchers) and [guides](https://redd.one/blog/practical-guide-to-custom-jest-matchers). There are a couple key problems, however. It is difficult or impossible to write the actual contents of errors to snapshot files--they are instead replaced with matcher names (see screenshot below). They are also somewhat inflexible, particularly when used with `jest-specific-snapshot`.
4. Wrapping expect in a try and rethrow. This [suggestion](https://stackoverflow.com/a/55097584) was made, and it would be a bit more general for the purposes of providing debug information, but it would require significant custom code in order to determine what the desired message to print would be. TaskFailures are the primary focus here.
5. Using built-in asymmetric property matchers. The final approach that I took is similar to option 3, but adds a bit more custom hacky code in order to ensure that as much useful information is written to the snapshot as possible. I use property matchers mapped to each individual taskFailure, and I apply them conditionally based on the test state and whether or not I have computed that failure to be present in the snapshot (see comments in `tests/utils/index.ts`). The only operational downside is that replacing existing snapshots with failure content becomes a bit more difficult, but not significantly.

### Follow-up Work
- I may consider also including debug information such as the trunkVersion and the linterVersion. If necessary, these wrappers can also be expanded to include other useful information through additional asymmetric matchers (custom or built-in).
- I may also integrate [retries](https://www.npmjs.com/package/jest-retries) if flaky failures are determined to be due to network issues after this PR brings out new insights.

Alternative 3
Screenshot 2023-01-21 at 11 45 52 PM


tests/utils/index.ts
```diff
@@ -215,3 +215,61 @@ export const skipOS = (excludedOS: string[]) => (_version?: string) =>
...
+    // Returning an any matcher for details, allowing for future assertions to succeed
+    return {
+      taskFailures: actual?.taskFailures?.map((failure) => ({
+        name: failure.name,
+        message: failure.message,
+        // trunk-ignore(eslint/@typescript-eslint/no-unsafe-assignment)
+        details: expect.any(String),
+      })),
+    };
+  }
```
TylerJang27, line 246: I favor a guard pattern here for easier context awareness, but if desired I can combine into the bottom return.

linters/eslint/test_data/eslint_v8.10.0_bad_install.check.shot
```diff
@@ -6,10 +6,12 @@ Object {
   "lintActions": Array [],
   "taskFailures": Array [
     Object {
+      "details": Any<String>,
       "message": "test_data/eof_autofix.ts... (4 files)",
       "name": "eslint",
     },
     Object {
+      "details": Any<String>,
```
det, line 14: Why use an any here? Can this be a discriminated union instead?

det (approved):

linters/eslint/test_data/eslint_v8.10.0_bad_install.check.shot
TylerJang27: Not necessary for https://jestjs.io/docs/snapshot-testing#property-matchers

tests/utils/index.ts
```diff
@@ -218,3 +218,61 @@ export const skipOS = (excludedOS: string[]) => (_version?: string) =>
...
+  };
+
+  return {
+    taskFailures:
+      actual?.taskFailures?.map((failure) => ({
+        name: failure.name,
+        message: failure.message,
+        // trunk-ignore(eslint/@typescript-eslint/no-unsafe-assignment)
+        details: snapshotContainsFailure(failure) ? expect.any(String) : failure.details,
+      })) ?? [],
```
dfrankland, line 276: small nit, optional: it's almost always better to default the inputs rather than the outputs for better clarity

```suggestion
      (actual?.taskFailures ?? []).map((failure) => ({
        name: failure.name,
        message: failure.message,
        // trunk-ignore(eslint/@typescript-eslint/no-unsafe-assignment)
        details: snapshotContainsFailure(failure) ? expect.any(String) : failure.details,
      })),
```

Same thing here with using `stringMatching`

tests/utils/landing_state.ts
```diff
@@ -13,11 +14,14 @@ const extractLintActionFields = ({
...
-  ...rest
-}: TaskFailure): TaskFailure => ({
+const extractTaskFailureFields = (
+  sandboxPath: string,
+  { detailPath: _detailPath, ...rest }: TaskFailure
+): TaskFailure => ({
   ...rest,
+  details: _detailPath
+    ? fs.readFileSync(path.resolve(sandboxPath, _detailPath), { encoding: "utf-8" })
+    : undefined,
```
dfrankland, line 24: seems like there's no reason to alias `detailPath` to `_detailPath` here since it gets used

dfrankland (approved):
all my comments are just nits really. code looks good to me though

PR #100, opened 2023-01-27
Parse Test Results For Uploading

TylerJang27:
Parses test results for uploading. Results end up in results.json file at the repo root during nightly runs.
- A known jest shortcoming is the inability to pass custom information from the test to its testResult/reporters. A hack that I've used to circumvent this is to hijack the console to report linter version information. This information is then filtered out and exfiltrated into the TestResult.
- I then added a new script to take the result from both OS runs from nightly and merge the test results down into a clearer synthesis of validated versions. The output results have the following invariant:
  - All linter/version pairs must have only passes and skips (and must have at least one pass). No fails are allowed.
  - The version on all passes must be identical. If there is a version mismatch, the pair will be excluded.

Remaining work includes:
- Iron out what ref we want to run these nightly tests on. We don't want to recommend versions that require newer config/fixes that users won't have yet, so probably only want uploads to come from the latest tagged ref.
- Add a cli integration to actually upload the results.
- Handle version mismatches that might be intentional from OS-varying downloads.

Sample output is viewable [here](https://github.com/trunk-io/plugins/actions/runs/4049695139/jobs/6966686224).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -30,5 +30,9 @@
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */,
     "traceResolution": false /* Enable tracing of module resolution */
   },
+  "ts-node": {
+    // Do not forget to `npm i -D tsconfig-paths`
+    "require": ["tsconfig-paths/register"]
+  },
   "exclude": ["node_modules"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,37 +1,29 @@
 {
   "compilerOptions": {
-    /* Visit https://aka.ms/tsconfig.json to read more about this file */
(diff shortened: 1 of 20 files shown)


tsconfig.json
```diff
@@ -1,34 +1,30 @@
 {
   "compilerOptions": {
-    /* Visit https://aka.ms/tsconfig.json to read more about this file */
```
TylerJang27, line 3: The transpilation for the reporter requires this to be a strictly formatted json file, so I've removed the comments here.

det (approved):

TylerJang27:
Added https://github.com/ansible/ansible-lint/issues/2969 for ansible-lint regression

PR #103, opened 2023-01-30
Support interactive hooks for trunk-fmt-pre-commit

TylerJang27:
Migrates this fix from the binary. This should enable users to run trunk-fmt-pre-commit from invocations such as VSCode's built-in SCM.

### DIFF
--- a/plugin.yaml
+++ b/plugin.yaml
@@ -1,5 +1,5 @@
 version: 0.1
-required_trunk_version: ">1.3.2-beta.7"
+required_trunk_version: ">=1.4.0"
 lint:
   files:
     - name: dml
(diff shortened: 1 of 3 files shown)


plugin.yaml
```diff
@@ -1,5 +1,5 @@
 version: 0.1
-required_trunk_version: ">1.3.2-beta.7"
+required_trunk_version: ">=1.4.0"
```
det:
```suggestion
required_trunk_version: ">1.4.0"
```

We required an unreleased trunk for shfmt.

det (approved):
lgtm, with that 1 change.

PR #107, opened 2023-02-02
Add action to automatically enable .git-blame-ignore-revs

det:


actions/git-blame-ignore-revs/plugin.yaml
```diff
@@ -0,0 +1,10 @@
+version: 0.1
+actions:
+  definitions:
+    - id: git-blame-ignore-revs
+      display_name: Auto-enable .git-blame-ignore-revs
+      description: Automatically configure git to use .git-blame-ignore-revs while the file exists.
+      run: ${cwd}/update_config.sh
+      triggers:
+        - files: [.git-blame-ignore-revs]
+        - schedule: 24h
```
dapirian, line 10: Why the schedule?

dapirian (approved):
Nice, also add to the top-level readme

actions/git-blame-ignore-revs/plugin.yaml
det: Just to cause it to run on startup. We need a `startup` trigger.

PR #139, opened 2023-02-05
Stop failing the annotate PR workflow when not running on forks

det:


EliSchleifer (approved):

.github/workflows/annotate_pr.yaml
```diff
@@ -21,3 +21,5 @@ jobs:
         uses: trunk-io/trunk-action@9cf65e08e822e9842fd9ef7ed2a2bd9092de0986 # v1.0.6
         with:
           post-annotations: true
+        # This job may fail when the PR was not run on a fork.
```
EliSchleifer, line 24: @det - can't we make this action not fail when not running on a fork
This is our own action isn't it

EliSchleifer:
I think the plan is to solve this correctly in the github action itself

det:
Let's land this for now, but still solve it in the action soon.

PR #147, opened 2023-02-07
Don't run latest tests on PRs

TylerJang27:
After this:
- PRs will only run on KnownGoodVersion (quick)
- Nightlies will run on Snapshots (slower) and Latest (non-idempotent due to releases)

This should make it so people don't become blocked by new releases. We can follow-up about where more coverage might be needed.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -35,7 +35,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        linter-version: [KnownGoodVersion]
         os: [ubuntu-latest, macos-latest]
     steps:
       - name: Checkout
@@ -51,7 +50,7 @@ jobs:
       - name: Linter Tests
         uses: ./.github/actions/linter_tests
         with:
-          linter-version: ${{ matrix.linter-version }}
+          linter-version: KnownGoodVersion
 
   # Run repo healthcheck tests
   repo_tests:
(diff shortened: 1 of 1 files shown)


EliSchleifer:
How does this impact snapshot execution. Seems like the only change here (which is good) is to only run the tests against well known version. Basically - does the existing system still work...which is a good test.

EliSchleifer (approved):
This change makes sense to me. Not sure I understand the snapshot semantics mentioned in the desciption

TylerJang27:
> How does this impact snapshot execution. Seems like the only change here (which is good) is to only run the tests against well known version. Basically - does the existing system still work...which is a good test.

@EliSchleifer This works fine in the context of snapshot execution. Everything (including nightlies) should work as is. Essentially the 3 tiers of snapshot tests are:
1. Snapshots (stable, may take longer because sometimes 2x-5x the tests per linter)
2. KnownGoodVersion (stable, relatively quick because usually 1 test per linter). This is strictly a subset of Snapshots.
3. Latest (semi-stable, may take longer due to downloading a newly released version)

Nightlies will continue to run 1 and 3 (and implicitly 2). PRs will just run 2 (we could run 1, but I think KGV is fine). We may lose a bit of coverage when people add new linters (that work on KGV but are broken on latest), but I have a follow-up that intends to address this.

TylerJang27:
@EliSchleifer one clarification. As is, KGV is the baseline/earliest known_good_version.

1.2.3 (latest)
snapshot 1.2.2
snapshot 1.2.1
snapshot 1.1.0 (known_good_version)

We can easily add an option to run on latest_good_version or similar, but this is how it is today

PR #150, opened 2023-02-07
Fix remark-lint tests

TylerJang27:
The last commit broke some of the tests. Namely, the no-tabs plugin was removed, the subcommands were renamed, and I believe the original file was saved with spaces.


EliSchleifer (approved):
LGTM

PR #154, opened 2023-02-10
Upload validated versions to staging and prod

TylerJang27:
Upload test result validated versions to both staging and prod. The daemon must be restarted before each upload, since the environment variable must propagate.

This will require a secrets change.


TylerJang27:
Need to chase down that upload error

det (approved):

PR #159, opened 2023-02-10
Add Slack Notifications for Test and Upload Failures

TylerJang27:
Adds Slack notifications to nightly runs with release tags. A couple notes:
1. This runs on release tags, so if we fix something or a definition on `main` we will still get notifications until we put out a new release.
2. Uses the test configuration from main, but the plugin definitions and individual tests from the latest release. This is to ensure compatibility with the workflow itself.
3. This works by having the `parse` script output a `failures.json` (in addition to the existing `results.json`), which includes a payload used to upload to Slack notifications. These json files are logged for ease of debugging, since they are not sensitive and dramatically improve the workflow investigation experience. In total, the following conditions will lead to a notification:
 a. A test failure occurs when a linter is run.
 b. Tests pass, but they run on different versions of a linter.
 c. Staging `trunk upload-linter-versions` fails.
 d. Prod `trunk upload-linter-versions` fails.

Also fixes a bug I had where skipped tests were resulting in "mismatched versions" and therefore no uploaded validated versions.

Example of [Upload Failure](https://github.com/trunk-io/plugins/actions/runs/4153694309/jobs/7185681558) (text API requirement has since been fixed) and [Normal Upload For Test Failure](https://github.com/trunk-io/plugins/actions/runs/4159123872/jobs/7195143382).


TylerJang27:
The test failures on Release are expected (ansible-lint `known_bad_versions` and nixpkgs-fmt needs newer rust runtime).

.github/workflows/nightly.yaml
```diff
@@ -141,6 +141,8 @@ jobs:
     # Still run on test failure
     if: always()
     timeout-minutes: 10
+    env:
+      SLACK_CHANNEL_ID: C04P1V21QF7
```
sxlijin, line 145: looking at the other examples, i'm pretty sure this can be a human-readable name

.github/workflows/nightly.yaml
```diff
@@ -197,6 +209,58 @@ jobs:
           trunk daemon shutdown
           trunk upload-linter-versions --token="$TRUNK_API_TOKEN" results.json
 
+      # Slack notifications
+      - name: Slack Notification For Failures
+        uses: slackapi/slack-github-action@v1
+        if: ${{ always() && steps.parse.outputs.failures == 'true' }}
```
sxlijin, line 215:
```suggestion
        if: always() && steps.parse.outputs.failures == 'true'
```

sxlijin (approved):

PR #161, opened 2023-02-13
Add default clang-tidy config

TylerJang27:
Use [suggested default config](https://github.com/trunk-io/configs/blob/main/.clang-tidy) whenever anyone runs `trunk check enable clang-tidy`. This will not auto-suggest/auto-init clang-tidy unless its config already exists.

I kept the original test config in, since it's a bit more particular and useful to have a bit of redundant test coverage here since the preCheck setup is so unusual.


linters/clang-tidy/.clang-tidy
```diff
@@ -0,0 +1,39 @@
...
+  -readability-magic-numbers,
+
+CheckOptions:
+  - key: readability-function-cognitive-complexity.Threshold
+    value: 100
+  - key: readability-function-cognitive-complexity.IgnoreMacros
+    value: true
+  # Set naming conventions for your style below (there are dozens of naming settings possible):
+  # See https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html
+  - key: readability-identifier-naming.ClassCase
```
det, line 32: We should delete our comment out the naming items.

det (approved):

PR #165, opened 2023-02-14
Add ruff linter support

tasgon:
Adds support for [ruff](https://github.com/charliermarsh/ruff), a Python linter written in Rust.


linters/ruff/plugin.yaml
```diff
@@ -0,0 +1,26 @@
...
+          output: sarif
+          parser:
+            runtime: python
+            run: ${cwd}/ruff_to_sarif.py
+          batch: true
+          success_codes: [0, 1]
+      runtime: python
+      package: ruff
+      direct_configs: [pyproject.toml]
+      affects_cache:
```
TylerJang27, line 18: An additional question I meant to ask earlier: does the ruff_cache/ folder actually affect ruff's cache? If so, we'll want to list it here so that it gets copied into the execution tree (not sure if this is sandboxed or not by default)

linters/ruff/plugin.yaml
```diff
@@ -0,0 +1,26 @@
...
+        - setup.cfg
+        - tox.ini
+      known_good_version: 0.0.243
+      is_recommended: false
+      good_without_config: true
```
TylerJang27, line 23: I thought you mentioned there was a default config you wanted to provide. Is that not the case?

TylerJang27 (commented):
Please reply to comments and add a PR description, otherwise looks mostly good.

linters/ruff/plugin.yaml
tasgon: As far as I can tell, it's not necessary. [Ruff will use a default config if a pyproject.toml isn't specified](https://github.com/charliermarsh/ruff#configuration).

TylerJang27: Ok. Sounds good then.

linters/ruff/plugin.yaml
tasgon: Should be resolved with the command changes.

TylerJang27 (approved):

PR #172, opened 2023-02-21
Fix ruff autofix parsing

tasgon:
Fixes a typo in the parser script and adds a workaround off-by-one error in ruff's output column locations in `fix` objects.

### DIFF
--- a/linters/ruff/ruff_to_sarif.py
+++ b/linters/ruff/ruff_to_sarif.py
@@ -9,12 +9,12 @@ results = []
 def get_region(entry):
     location = entry["location"]
     region = {
-        "startColumn": location["column"],
+        "startColumn": location["column"] + 1,
         "startLine": location["row"],
     }
     if "end_location" in entry:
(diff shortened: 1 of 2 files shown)


linters/ruff/ruff_to_sarif.py
```diff
@@ -6,15 +6,15 @@
 results = []
 
 
-def get_region(entry):
```
TylerJang27, line 9: The one piece I was missing--is this issue you were describing just specific to ruff/to this parser? If so, can you add a comment before line 59 and maybe before line 9 briefly describing the reason for this off-by-one fix?

TylerJang27 (commented):
One comment, otherwise good

TylerJang27 (approved):
LGTM

linters/ruff/ruff_to_sarif.py
tasgon: This is something specific to Ruff. Just added comments

PR #173, opened 2023-02-22
Extend ruff support for jupyter files

siddharthab:
`ruff` does not support jupyter files natively, so we need to run it
through nbqa which extracts the code blocks and then runs `ruff`.

Also test that `black` works on jupyter files.


siddharthab:
Because the issues are generated from a temporary file, the line number annotations are not correct. I am not sure if there is a way to map these line numbers to the original lines in the .ipynb file.

siddharthab:
cc @patrick-kidger

siddharthab:
cc @tasgon

TylerJang27:
Hi there! Thanks for using trunk and for making this contribution! We're happy to see our ruff support getting some traction, and thanks for helping us iron out the kinks.

At first glance this PR and the testing changes look great--thanks for modifying the testing setup when appropriate. I will give this a more thorough review tomorrow morning, and I'll talk to the team about how we might address the line number issue and nbqa to make sure our diagnostics are correct. Thanks!

siddharthab:
Apparently nbqa does do the reverse mapping for `black` [here](https://github.com/nbQA-dev/nbQA/blob/main/nbqa/output_parser.py). So we might just have to add functionality in nbqa for ruff json output.

siddharthab:
The mapping that nbqa will give us will be line number inside a notebook cell. I am not sure how to generate a SARIF for GitHub to annotate on the rendered notebook. I am not sure if this is even possible.

Because in-line annotations for notebooks is a bigger issue, I am going to change this PR to make the new linter a pass-fail linter.

det (approved):
This looks great! Thank you for your contribution!

PR #175, opened 2023-02-24
Use a default config for sqlfluff so users can override the dialect

det:
When we hard-code it on the command-line, the config file has no affect on the dialect.
Add a default config file as without a config file specifying a dialect the linter will fail.

Upgrade everything too, thanks LUV.


TylerJang27 (approved):
LGTM

PR #176, opened 2023-02-24
Migrate over file types

TylerJang27:
Migrates file types and comment formats from their original definition in the binary. These are stored in roughly alphabetical order in `linters/plugin.yaml`.

Comment formats won't actually be meaningfully loaded until the next cli release, but having them here won't break anything.


det:
Any thoughts on splitting these up into `files/<name>/plugin.yaml` ?

TylerJang27:
> Any thoughts on splitting these up into `files/<name>/plugin.yaml` ?

@det I did consider doing something like `files/plugin.yaml` and `comments/plugin.yaml` I think they just become too annoying to manage, and given that some of them inherit from others, I think it's just universally easier to have them in one place. In addition, with the directory structure of the plugins repo, I really think they should be inside of `linters/` (rather than at root since it has nothing to do with actions), but it doesn't make sense for them to be within another subdirectory inside of that, for fear it looks like a "files" linter. That's why I landed on where I have it right now, I think it stands out.

det (approved):
lgtm

PR #177, opened 2023-02-27
Optional interactivity for trunk-check-pre-push

siddharthab:
Following from discussion in the [Slack thread](https://trunkcommunity.slack.com/archives/C0205B6KK8X/p1677368211050039).


det (commented):
Thanks for the contribution! Unfortunately, when pushing from vs code the experience for pre-push hooks is not great.
1) There is no progress while the hook is running.
2) When it is blocked due to a lint issue it does not give an opportunity to override the tool and push anyways.
3) When it is blocked due to a lint issue it just says `failed to push some refs ...` and doesn't properly forward the error to the user in any way.

Due to these issues, we would prefer to not make this the default. Could we instead duplicate the hook to `trunk-check-pre-push-always` so people can opt-in to this behavior if they like?

siddharthab:
Yes, I think that's a good idea. I can document in our repo that VS Code and `gh pr create` users should expect this behavior.

det (approved):

PR #178, opened 2023-02-27
Add missing ktlint download and repo test

TylerJang27:
Added ktlint download that I missed when I originally ported definitions over and another of its tests. I also added a repo test to make sure this doesn't happen again.

### DIFF
--- a/tests/repo_tests/no_missing_download.test.ts
+++ b/tests/repo_tests/no_missing_download.test.ts
@@ -21,6 +21,7 @@ describe("All linters that use downloads must define them", () => {
       // trunk-ignore(eslint/jest/valid-title)
       it(linter, () => {
         // trunk-ignore-begin(eslint): Expected any accesses
+        // Ignoring no-unsafe-member-access, no-unsafe-assignment, no-unsafe-call, no-unsafe-return, and conditional jest expect
         const yamlContents = parseYaml(path.resolve(linterDir, linter, "plugin.yaml"));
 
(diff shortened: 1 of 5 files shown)


tests/repo_tests/no_missing_download.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  const linters = fs
+    .readdirSync(linterDir)
+    .filter((file) => fs.lstatSync(path.resolve(linterDir, file)).isDirectory());
+
+  linters
+    .filter((linter) => !excludedLinters.includes(linter))
+    .forEach((linter) => {
+      // trunk-ignore(eslint/jest/valid-title)
+      it(linter, () => {
+        // trunk-ignore-begin(eslint): Expected any accesses
```
sxlijin, line 23: which eslint errors are you ignoring with this?

sxlijin (approved):

tests/repo_tests/no_missing_download.test.ts
TylerJang27: 1 jest warning about conditional expects and 
- no-unsafe-member-access
- no-unsafe-assignment
- no-unsafe-call
- no-unsafe-return

I can add to the comment for this. Some of our ignore code complains about line numbers, so I didn't do it in the first place

sxlijin: just add it to a comment below or above the begin

PR #179, opened 2023-02-28
Fix run command for trunk-announce

TylerJang27:
Includes a fix to make sure pre-rebase hooks are handled correctly.


laurit17 (approved):
LGTM - though I'm curious how this bug came to be originally - `default_actions` has the correct run command:

https://github.com/trunk-io/trunk/blob/main/trunk/cli/actions/data/default_actions.yaml#L5

TylerJang27:
> LGTM - though I'm curious how this bug came to be originally - `default_actions` has the correct run command:
> 
> https://github.com/trunk-io/trunk/blob/main/trunk/cli/actions/data/default_actions.yaml#L5

It never got copied over, looks like

PR #182, opened 2023-03-02
Shellcheck fix

det:


TylerJang27 (approved):
Should follow this up with a test for following `source`

PR #183, opened 2023-03-02
Add tests for old plugin linters

TylerJang27:
Adds tests for linters that were added before the testing framework was built. They should now be included in the Linter Upgrade Validation logic, and the rest of our repo health checks. The diff is mostly snapshots.

- sqlfmt
- circleci: plugin config was formerly broken for pass_fail output
- codespell
- cspell
- nancy: plugin config was formerly broken, in partial due to recent target changes. **This also requires a newer trunk version with a bugfix for the enable command to pass nancy tests and be merged.**
- oxipng: snapshots don't work well for this, but added an additional assertion

We should probably follow this up with some work to get git-lfs working properly for plugins.


TylerJang27:
Nancy tests are having issues on Mac. Investigating

sxlijin (approved):

PR #185, opened 2023-03-04
Auto-init with yamllint and ruff

TylerJang27:
Changes auto-init behavior as follows:
- Adds yamllint during auto-init with a default config.
- Adds ruff during auto-init with a default config. I've attempted to model this closely off our default flake8 config and using ruff [rules](https://beta.ruff.rs/docs/rules/).
- No longer auto-enable flake8 (avoid conflicting python linters).


dapirian (approved):
lgtm!

PR #196, opened 2023-03-09
Improve versioning and installs for commitlint

TylerJang27:
Fixes commitlint to at least have a default config specification (originally it errored, complaining about the lack of one).

Also provide instructions on how to override the listed commitlint plugins (the recommendation requires >1.6.0, hence the specification. I can make this >=1.6.0 if desired).

In the future, we will probably want to explicitly version these sort of actions via config so that we can guarantee hermetic treatment.


laurit17 (commented):
I recall some discussion about what might enforce the consistency of these versions - like an action - is that captured somewhere?

TylerJang27:
> I recall some discussion about what might enforce the consistency of these versions - like an action - is that captured somewhere?

@laurit17 Overriding the packages file and checking it in could achieve this... otherwise I'm not sure what you're referring to?

TylerJang27:
Filed https://linear.app/trunk/issue/TRUNK-6288/support-a-versioning-story-for-actions, we can triage as needed during our next planning

laurit17 (approved):

PR #198, opened 2023-03-09
Support latest jest-specific-snapshot

TylerJang27:
#194 broke all of our snapshots, since config was changed to respect the [new defaults](https://github.com/igor-dv/jest-specific-snapshot/releases/tag/v8.0.0). We could override this through config, but it's not trivial to wire through due to our custom `jest-specific-snapshot` handling, so we might as well support the new, more concise default anyway.

Also adds a test filter so this sort of things can't be auto-merged in the future. I eyeball-audited all of the updates to make sure they weren't broken.

Snapshots updates by running `PLUGINS_TEST_LINTER_VERSION=Snapshot npm test linters -- -u`

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -8,6 +8,7 @@ all-linters:
   - tests/utils/**
   - tests/index.ts
   - .trunk/trunk.yaml
+  - package.json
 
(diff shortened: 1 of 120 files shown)


dapirian (approved):
Lgtm thanks 🙏

PR #199, opened 2023-03-10
Added tfsec as a security linter

puzzler7:
Added tfsec as a security linter. Includes a small fix to testing that prevents accidentally linting the json output of trunk check.


sxlijin (commented):
LGTM but why are we deleting package-lock.json?

puzzler7:
Re: package-lock.json: I'm not sure - when I initially made the PR, CI told me to run npm install, and I did and pushed the result. I don't know what the implications of that are.

Re: .json -> .out.json: tfsec runs on terraform files, which are `.tf` and `.tf.json` files. Additionally, it runs on the parent directory, rather than an individual file. When running the tests before this change, the test harness would run tfsec on the file `aws.in.tf`, and write the output to `aws.in.tf.json` in the same directory, at which point tfsec would attempt to run on the newly created file and error.

tests/driver/index.ts
```diff
@@ -453,7 +453,7 @@ export class TrunkDriver {
    */
   async runCheckUnit(targetRelativePath: string, linter: string): Promise<TestResult> {
     const targetAbsPath = path.resolve(this.sandboxPath ?? "", targetRelativePath);
-    const resultJsonPath = `${targetAbsPath}.json`;
+    const resultJsonPath = `${targetAbsPath}.out.json`;
```
sxlijin, line 456: add comment here explaining the tf.json thing

sxlijin (approved):
Let's back out the package-lock.json changes - I suspect you ran into some kind of transient error, because we haven't modified package.json.

PR #200, opened 2023-03-10
Truncate slack notifications

TylerJang27:
The API is limited to 50 blocks, so if there are more than 50, change the 50th to be "and N more".

This was the bug that caused us not to get a notification for the failures last night (which were caused by the update to `jest-specific-snapshot` on main being mismatched with the snapshots on release). Tested the parse result generation locally.


sxlijin (approved):

dapirian (approved):

TylerJang27:
Will merge this after #183

PR #203, opened 2023-03-12
Integrate checkov linter

laurit17:
TODOs:
- this supports a number of additional config file types - kubernetes, helm, azure - however, not clear how to distinguish these file types. - the yaml has a general structure for which we can perhaps create a file type config, but it is not particularly unique. I suppose users can set up their own file type?


.trunk/trunk.yaml
```diff
@@ -16,6 +16,7 @@ runtimes:
 
 lint:
   enabled:
+    - checkov@2.3.75
```
TylerJang27, line 19: We don't need this enabled ourselves do we?

linters/checkov/plugin.yaml
```diff
@@ -0,0 +1,17 @@
+version: 0.1
+lint:
+  definitions:
+    - name: checkov
+      files: [terraform, cloudformation, docker]
+      runtime: python
+      package: checkov
+      commands:
+        - name: lint
+          run: checkov -f ${target} -o sarif --output-file-path ${tmpfile},
```
TylerJang27, line 10: comma typo?

TylerJang27 (commented):
Looks mostly good. Let's sync up on Monday about file types and a couple other things

.trunk/trunk.yaml
laurit17: no, removed

linters/checkov/plugin.yaml
laurit17: no - it is intentional

PR #204, opened 2023-03-14
fix(clippy): allow updating the local cache

sxlijin:
`--frozen` forces clippy to not update the local cache, but this breaks clippy in CI scenarios (the original reason that this was OK was that users would populate this cache themselves)

`--locked` asserts that the lockfile stays locked, which is what we really want for linter invocations


TylerJang27 (approved):
LGTM, good catch

PR #205, opened 2023-03-14
Added trivy config subcommand

puzzler7:
Added trivy config as a linter command.

I've added a parser from json to sarif, rather than using the native sarif output, because the sarif output generates messages such as the one below, literal newlines and all.

```
Artifact: basic.in.yaml
Type: kubernetes
Vulnerability KSV106
Severity: LOW
Message: container should drop all
Link: [KSV106](https://avd.aquasec.com/misconfig/ksv106)
```


.trunk/trunk.yaml
```diff
@@ -16,6 +16,7 @@ runtimes:
 
 lint:
   enabled:
+    - trivy@0.38.2
```
TylerJang27: Probs want to back this out

linters/trivy/plugin.yaml
```diff
@@ -16,7 +16,7 @@ lint:
 
   definitions:
     - name: trivy
-      files: [lockfile]
+      files: [lockfile, docker, yaml, terraform]
```
TylerJang27, line 19: Just to sanity check, trivy config is a no-op on lockfiles/shouldn't break anything/be noisy, right?

linters/trivy/plugin.yaml
```diff
@@ -29,7 +29,18 @@ lint:
           batch: false
           parser:
             runtime: python
-            run: ${plugin}/linters/trivy/trivy_to_sarif.py
+            run: ${plugin}/linters/trivy/trivy_fs_to_sarif.py
+
+        - name: config
+          output: sarif
+          run: trivy config ${target} --format json --cache-dir ${shared_cachedir}
```
TylerJang27, line 36: Does this need a --no-progress?

linters/trivy/test_data/basic.in.yaml
TylerJang27, line 1: So because this is a `customLinterCheckTest`, these don't need the `basic.in.__` file name format. It's not the biggest thing, but I'd prefer if you changed the names of these to shorter ones/Dockerfile just to keep convention. (and regenerate the snapshot)

TylerJang27 (approved):
Looks good. Couple things just to manually verify/follow-up, and then can merge.

Please create a ticket if you haven't to remind about the `is_security` work

sxlijin (approved):

PR #208, opened 2023-03-15
Added osv-scanner

puzzler7:
Added osv-scanner as a linter. I added a new filetype osvlockfile (after talking with Tyler), because osv-scanner does not support go.sum files.


linters/osv-scanner/plugin.yaml
```diff
@@ -0,0 +1,34 @@
...
+        - os:
+            linux: linux
+            macos: darwin
+          cpu:
+            x86_64: amd64
+            arm_64: arm64
+          url: https://github.com/google/osv-scanner/releases/download/v${version}/osv-scanner_${version}_${os}_${cpu}
+  definitions:
+    - name: osv-scanner
+      files: [osvlockfile]
```
TylerJang27, line 17: nit: change name of file type to osv-lockfiles. I read current as a lockfile used for osv

linters/osv-scanner/plugin.yaml
```diff
@@ -0,0 +1,34 @@
...
+          success_codes: [0, 1, 128]
+          read_output_from: stdout
+          parser:
+            runtime: python
+            run: ${plugin}/linters/osv-scanner/osv_to_sarif.py
+      environment:
+        - name: PATH
+          list: ["${linter}"]
+      version_command:
+        parse_regex: "version: ${version}\n"
```
TylerJang27, line 33: nit: try locally to see if you can remove \n. Our semver here might be a line rather than full match

linters/plugin.yaml
```diff
@@ -310,6 +310,55 @@ lint:
       extensions:
         - m
 
+    # This is the same as the lockfile type above, minus go.sum because osv-scanner does not support that
+    # TODO: TYLER once linter ignores are supported, move this into a linter ignore
```
TylerJang27, line 314: nit: TODO(Tyler)-just for consistency and so I can find it more easily

TylerJang27 (commented):
Looks good. Gonna leave a chance for Sam to scan it

linters/plugin.yaml
```diff
@@ -310,6 +310,55 @@ lint:
       extensions:
         - m
 
+    # This is the same as the lockfile type above, minus go.sum because osv-scanner does not support that
+    # TODO(Tyler): once linter ignores are supported, move this into a linter ignore
```
sxlijin, line 314: alternatively: can we say osv applies to all lockfiles, and then inside default_lint.yaml ship a default ignore for go.sum / osv-scanner?

sxlijin (approved):
hurrah! excited to get this in :)))

linters/plugin.yaml
TylerJang27: Discussed this with Mav, I think the best approach here is to:
1. Land this (sorry I saw this comment after I merged, but compat/history shouldn't be a big obstacle)
2. Add the logic in cli to support merging ignores from the plugin repos, and then add that ignore (and all the rest) here (since it's specific to osv)
  a. One complexity is determining a merge key, since ignores don't have names, etc. Maybe we ship our default set and just add the option to append ignores.
3. Alternatively to #2, just do as you mention and ship the default_lint with a new osv-scanner ignore
4. Regardless, at that point we can ditch this filetype from the plugins repo and bump our `required_cli_version`

PR #210, opened 2023-03-15
Add gofumpt

tomlazar:
Basically the same as gofmt, just a little more strict.


linters/gofumpt/gofumpt.test.ts
```diff
@@ -0,0 +1,6 @@
+import { linterCheckTest, linterFmtTest } from "tests";
+
+// gofumpt generates a failure on an empty file
+linterCheckTest({ linterName: "gofumpt", namedTestPrefixes: ["empty"] });
```
TylerJang27, line 4: Thanks for calling this out. We may move in the future to adding `allow_empty_files: false` to both gofumpt and gofmt, but this should be fine for now.

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "plugins",
+  "name": "trunk-plugins",
```
TylerJang27, line 2: Can you back this file change out? Was there something you ran into that brought this up?

TylerJang27 (approved):
Hi there! Thanks for using trunk and for making this contribution! The setup for gofumpt looks great, with just a couple minor comments.

Once you've addressed those feel free to update your PR with main and merge! Thanks again!

package-lock.json
tomlazar: nothing specific, must have happened since I renamed my fork as `trunk-plugins`

tomlazar:
@TylerJang27 should be all good to go!

TylerJang27:
@tomlazar Thanks again! We probably won't get another release out for a week or so, but if you need `gofumpt` sooner, you can change the plugin source in your trunk.yaml to look like:

```yaml
plugins:
  sources:
    - id: trunk
      ref: 19526b8797860900fb9680233943072b78c9f356
      uri: https://github.com/trunk-io/plugins
```

PR #211, opened 2023-03-16
Added is_security to all security linters

puzzler7:
Added is_security to all security linters, and updated their snapshots. Additionally bumped the cli version so that is_security works.


TylerJang27 (approved):
LGTM, just check with Sam on scoping to make sure this is ready to land

sxlijin (approved):
🙌

PR #219, opened 2023-03-17
Add action trunk-check-pre-commit

TylerJang27:
This has been requested several times by community users. It's generally not preferable over `trunk-fmt-pre-commit` since linters can be quite slow. I've added instructions via comments for how to apply autofixes. Users can also use --filter/--exclude if they want.

I will personally probably turn this on for this repo in my user.yaml for eslint with autofixes enabled.

### DIFF
--- a/actions/trunk/plugin.yaml
+++ b/actions/trunk/plugin.yaml
@@ -26,6 +26,19 @@ actions:
         - git_hooks: [pre-commit]
       notify_on_error: false
 
+      # We do not auto-suggest turning on check pre-commit for several reasons:
+      # - Linters can often be much slower than formatters, and we don't want to unnecessarily block iteration.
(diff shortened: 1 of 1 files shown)


actions/trunk/plugin.yaml
```diff
@@ -26,6 +26,19 @@ actions:
         - git_hooks: [pre-commit]
       notify_on_error: false
 
+      # We do not auto-suggest turning on check pre-commit for several reasons:
+      # - Linters can often be much slower than formatters, and we don't want to unnecessarily block iteration.
+      # - The trunk-fmt-pre-commit action is usually sufficient for enforcing style.
```
det, line 31:
```suggestion
      # - The trunk-fmt-pre-commit action is usually sufficient for enforcing style.
      # - It's often desired to temporarily commit code that has linter issues.
```

det (approved):
lgtm with comment

PR #223, opened 2023-03-21
Add support for python 3.11 runtime

fergusdixon:
The latest release added support for 3.11.1 
https://github.com/indygreg/python-build-standalone/releases/tag/20230116


TylerJang27 (approved):
Hi @fergusdixon, thanks for the PR! The runtime addition looks great! Note that we're not bumping the runtime `known_good_version` yet, so you will need to specify `python@3.11.1` in your `.trunk/trunk.yaml`'s runtime.enabled section for it to be picked up, once we make our next plugins release.

PR #224, opened 2023-03-22
feat: add default config for stylelint

sxlijin:
Also make stylelint work with scss


linters/stylelint/test_data/basic.in.scss
TylerJang27, line 1: For [lack of forethought and for simplicity] reasons, all test input files must have different prefixes, even if they have different file extensions. Please rename one of these. I usually prefix as basic_py, or similar. You'll notice that none of these snapshots contain "basic.in.sass" for this reason.

linters/stylelint/plugin.yaml
```diff
@@ -1,5 +1,15 @@
...
+  # When enabling stylelint, you should also add the stylelint plugin configs,
+  # like so:
+  #
+  #   lint:
+  #   enabled:
+  #     - stylelint@15.3.0:
+  #         packages:
+  #           - stylelint-config-standard@31.0.0
+  #           - stylelint-config-standard-scss@7.0.1
+  #
```
TylerJang27, line 12: I understand that this is strictly an improvement, but I'm not sure if this is discoverable enough. Thoughts on moving this to a `linters/stylelint`-level readme or perhaps somewhere else?

linters/stylelint/plugin.yaml
```diff
@@ -1,5 +1,15 @@
...
   definitions:
```
TylerJang27, line 13: Because you've added a direct config, you must set `is_recommended: false` (this is causing the repo_tests failure)

TylerJang27 (commented):
Couple notes, mainly style choices and testing semantics. Also looks like you have a couple tests failing, not immediately sure why but shouldn't be too hard to reason about. Ping me when ready for re-review

PR #225, opened 2023-03-22
Migrate over default lint ignores, environments

TylerJang27:
Migrates these over, which gives us greater flexibility with security linters, among other things. This PR will be a no-op until the next release of the CLI, since these are not yet merged from sourced plugins.

In total, handles the following:
- Adds default lint ignores
- Adds default environment(s)
- Removes `trunk-cache-prune` definition (will be internal still)
- Fixes a small bug where the change set of this PR was triggering all latest linter tests to run


sxlijin (approved):
`failure() || success()` feels like it could be `always()` - but fix that separately

TylerJang27:
> `failure() || success()` feels like it could be `always()` - but fix that separately

failure() and success() do not cover the full set of options (always() is more encompassing than I want. Can talk more offline)

PR #226, opened 2023-03-22
Add snapshot for new circleci output format

TylerJang27:
The output changed as of https://github.com/CircleCI-Public/circleci-cli/releases/tag/v0.1.24705 (see [failed job](https://github.com/trunk-io/plugins/actions/runs/4493495218/jobs/7906295231)), adding a new snapshot with a RELEASE tag should provide no interruption to the linter validation service.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -80,7 +80,7 @@ jobs:
           cp package-lock.json package-lock.json.bak
           cp -r .github/actions .github/actions.bak
           # Include any newly generated snapshots that have been marked release-ready
-          grep "// RELEASE" -r --include=*.shot -l | xargs -I {} cp {}{,.bak}
+          grep "// trunk-upgrade-validation:RELEASE" -r --include=*.shot -l | xargs -I {} cp {}{,.bak}
         continue-on-error: true
 
       - name: Get Latest Release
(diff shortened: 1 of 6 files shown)


TylerJang27:
For a bit of additional context on the `// RELEASE` tag, here's a brief design overview:

- Last week I added support for tagging snapshot (it could in theory be any file, but for now it's just snapshots for safety reasons) as `// RELEASE`.
- When preparing the tests for validating linter versions, a couple things happen:
  - The latest `tests/` dir is checked out (it must be backwards compatible for this reason)
  - `package.json` and `package-lock.json` are checked out (dependency of the above)
  - `.trunk/trunk.yaml` is checked out.
- These ensure that tests run effectively on the latest version of the Trunk cli and with the latest testing framework updates.
- In addition, any snapshots marked with `// RELEASE` will be checked out from main, which allows us to quickly adapt and mark validated support for newly released versions that only include small, non-breaking changes to the snapshots we test them against. This case for circleci, since it's just a pass-fail linter and doesn't require any complicated parsing, is one of those cases.
- There is a trunk action that can be run manually to remove all of those release tags after we make a release.

sxlijin:
can we make it so that it's a bit more obvious that `// RELEASE` has obvious semantics? e.g. `// trunk-luv:RELEASE`

sxlijin (approved):

PR #228, opened 2023-03-24
Fix script for trunk-check-pre-push-always

siddharthab:
The script currently assumes that trunk can be found on PATH. We can instead inject logic with a copy of the code that git hooks launcher script uses.

We should also exec the formed command line in the script.


TylerJang27 (approved):
Thanks for catching this. We'll be looking into providing this as a built-in fix as well in the near future.

PR #229, opened 2023-03-27
Add Poetry actions

EliSchleifer:
Adds `poetry` Trunk Actions to match parity with [`pre-commit`](https://python-poetry.org/docs/pre-commit-hooks/). Note that the non-hermetic story is much cleaner at time of writing, with the hermetic story blocked by not being able to unset `VIRTUAL_ENV` in the Action and Tool. In other words, for now recommended behavior requires manual [Poetry installation](https://python-poetry.org/docs/#installation).

We've added a tool test and several actions tests. PR taken up by @TylerJang27


EliSchleifer:
There is a bug we've identified that blocks a commit when all the poetry actions are enabled as defined. The trunk team is actively investigating.

PR #230, opened 2023-03-27
Simple python hello world action example

EliSchleifer:
1. Basic hello world example of python action triggered by pre-commit
2. Of note - the hello-world-python uses a requirements.txt file to bring in a printing library. This demonstrates the basic ability of plugins to support hermetic requirements per action.


TylerJang27 (approved):
Looks good. Only minor suggestion is maybe to add a readme inside with really basic instructions (e.g. "`trunk actions enable hello-world-python` to enable, then run with `trunk run hello-world-python` or during git-commit...")

PR #236, opened 2023-03-31
Added DJLint, a formatter for templated HTML files.

fsargent:
<details open="true"><summary>Generated summary (powered by <a href="https://app.graphite.dev">Graphite</a>)</summary>

> #### What Changed
> Added
> - DJLint linter
> - Test case for djlint
> - djlint configuration files
> #### Testing
> Run the test suite and verify that the djlint linter is working as expected.
> #### Reasoning
> To add support for the DJLint linter and improve the overall code quality of the project.
</details>


linters/djlint/djlint.test.ts
```diff
@@ -0,0 +1,4 @@
+import { linterCheckTest } from "tests";
+
+// TODO(Tyler): We will eventually need to add a couple more test cases involving failure modes and other packages.
+linterCheckTest({ linterName: "djlint" });
```
TylerJang27, line 4: Please rename `logout.html` -> `logout.in.html` in order for the test to pick it up (and then run `npm test djlint` locally to generate the snapshot, which acts as the expected output. For more info, see `tests/readme.md`

linters/djlint/README.md
```diff
@@ -0,0 +1,10 @@
+# DJLint
+
+- [README](https://github.com/Riverside-Healthcare/djlint#readme)
+- [Rules](https://www.djlint.com/docs/linter/)
+
+DJLint supports Django • Jinja • Nunjucks • Twig • Handlebars • Mustache • GoLang • Angular
+
+For use of anything other than Django, please visit the readme and set the appropriate profile.
```
TylerJang27, line 8: Good callout. Can you add "set the appropriate profile in your `run` command or in the linter config"?

As another note, in the future, we will likely add command-specific filetypes and hopefully we can do this sort of profile management automatically within the same linter.

linters/djlint/.djlintrc
```diff
@@ -0,0 +1,3 @@
+{
+  "linter_output_format": "{filename}:{line} - {code} - {message} :: {match}"
```
TylerJang27, line 2: Probably don't need {match} here, since it looks like all of these are optional. If you suspect it provides additional value, lmk. Otherwise, I really appreciate having this output format option.

TylerJang27 (commented):
Hi @fsargent, thanks for the PR. DJLint looks like a great formatter and the integration looks great! Can you make the testing changes? Assuming those pass on the next run, this should be almost ready to land.

linters/djlint/.djlintrc
fsargent: Match is super useful here as it describes what the error is talking about.
test_data/logout.html:7:4 - H017 - Tag should be self closing. <link>
 '<link>'  is the match for this example. 

Rendered in Trunk as 

```
**test_data/logout.html:7:4**
 7:4  high  Tag should be self closing. :: <link>  **djlint/H017**
```

linters/djlint/djlint.test.ts
TylerJang27, line 4: Can you also add a `linterFmtTest` and rerun the test to generate a snapshot? Sorry I should've caught this in the first pass. This should mirror sqlfluff test, but without the `preCheck` setup.

linters/djlint/plugin.yaml
```diff
@@ -0,0 +1,36 @@
...
+          in_place: true
+      direct_configs: [.djlintrc]
+
+      affects_cache:
+        - .flake8
+        - pyproject.toml
+        - setup.cfg
+        - tox.ini
+      issue_url_format: https://www.djlint.com/docs/linter/#rules
+      known_good_version: 1.19.16
```
TylerJang27, line 33: Please add `is_recommended: false` to the linter's definition. This is the root of the repo_test failure because it's causing djlint to be auto-enabled as is with its config. You can make sure the repo_tests are passing locally by running `npm test tests/`

TylerJang27 (commented):
2 final things, mainly just for convention and testing-compatibility.

linters/djlint/plugin.yaml
fsargent: Shouldn't it be recommended though?
I updated the PR which hopefully won't break things, but let me know if that's not ideal for how djlint works.

fsargent:
I think DJLint and Prettier might fight on how they format template files, and it's hard to detect when a HTML file is a template file or when it's raw HTML. Something to worry about!

linters/djlint/plugin.yaml
TylerJang27: Sorry I should've explained this a bit more. Right now the logic for whether trunk auto-enables a linter during `trunk init` (or auto-suggest it during `trunk upgrade`) essentially boils down to the following logic:
1. If the linter has a direct config in its definition folder (like you have here with .djlintrc) and `is_recommended: true (default)`, auto-enable.
2. If the linter has `good_without_config: true` and `is_recommended: true (default)`, auto-enable.
3. Otherwise, don't auto-enable.

We're planning to improve the naming and interaction of these fields, but as is, djlint would be auto-enabled for users. Usually we give linter integrations at least a couple weeks to marinate/stabilize before we auto-suggest them for all of our users. Especially if djlint and prettier might step on each other, that's another reason we should probably hold off.

So can you revert your `config_check_test.ts` change and add `is_recommended: false`? After dogfooding for a bit we can revisit auto-enabling it. Sorry for the confusion!

TylerJang27 (approved):
LGTM! Thanks for the PR!

PR #245, opened 2023-04-10
Make trufflehog batchable

TylerJang27:
I probably should have caught this in the first pass, but from what I can tell, there's no reason it shouldn't be batchable, and it frequently errors on medium/large codebases when not batched and running with `-a`.


puzzler7 (approved):
LGTM

PR #248, opened 2023-04-12
Update osv snapshot

TylerJang27:
Address [failure](https://github.com/trunk-io/plugins/actions/runs/4676037039) and mark ready for release. We've had osv-scanner integrated for a limited time, but if we see repeated things like this, we may want to consider a specialized assertion for it.


puzzler7 (approved):
LGTM

PR #250, opened 2023-04-14
Gokart

gejustin:
This adds support for [Gokart](https://github.com/praetorian-inc/gokart), a vulnerability scanning tool for Golang.


EliSchleifer:
@gejustin - Looks like the gokart tests didn't run correctly. Let us know if you need help digging into why the test is not working correctly. @TylerJang27 for viz.

linters/gokart/plugin.yaml
```diff
@@ -0,0 +1,30 @@
...
+      package: github.com/praetorian-inc/gokart
+      environment:
+        - name: PATH
+          list: ["${linter}"]
+      run_from_root_file: go.mod
+      commands:
+        - name: lint
+          output: sarif
+          target: ${parent}
+          run: gokart scan -i ${workspace}/analyzers.yml -s ${target}
```
TylerJang27, line 16: After investigation, you were correct. This should be:
`run: gokart scan -i analyzers.yml -s ${target}`

TylerJang27 (commented):
Thanks for the contribution! Just need to fix the run command, as you suspected.

TylerJang27 (approved):
LGTM. Thanks for the contribution!

PR #251, opened 2023-04-15
Add downloads and configs for windows

det:


TylerJang27 (approved):
LGTM. Your discretion on when we want to land and/or release this in conjunction with the gokart (merged, unreleased), and tools (unlanded) work

PR #252, opened 2023-04-17
feat: recognize .buildifier.json

sxlijin:
Buildifier configuration is not very well documented: source code comments provide some pointers to how this works, but for the most part you have to actually read the source code to figure out how to feed configuration in.

The semantics are this:

  * a `.buildifier.json` at the root of the workspace will be read into the configuration for a buildifier invocation ([link](https://github.com/bazelbuild/buildtools/blob/6.1.0/buildifier/buildifier.go#L85))

  * `tables` specifies a path to a tables config that _overwrites_ the default tables config

  * `add_tables` specifies a path to a tables config that is merged into the default tables config

  * tables configs allow overriding buildifier's behavior with respect to a given attribute ([link](https://github.com/bazelbuild/buildtools/blob/6.1.0/tables/tables.go)); for example, adding an attribute name `fizz` to `IsSortableListArg` tells buildifier that all `fizz` attributes for all rules should be sorted

Note that buildifier JSON parsing is lenient: unrecognized fields do _not_ throw an error (this is how the "human documentation" fields in the configs provided in this PR work).


linters/buildifier/buildifier.test.ts
```diff
@@ -1,5 +1,32 @@
...
+    "fizz": true
+  },
+  "SortableAllowlist": {},
+  "NamePriority": {}
+}
+  `;
+
+  driver.moveFile(".buildifier.json", ".trunk/configs/.buildifier.json");
+  driver.deleteFile(".buildifier-tables.json");
+  driver.writeFile(".trunk/configs/.buildifier-tables.json", buildifierTablesContents);
```
sxlijin, line 21: Not sure how this should be adjusted: should this test go via `.buildifier-tables.json` instead?

linters/buildifier/.buildifier.json
TylerJang27, line 1: Are you intending for this file  and .`buildifier-tables.json` to be included in users' repos on creation/dumped?

If yes:
1. Rn dumping direct configs is first-match (max of one config will be dumped). Presumably this is deterministic but idk if it's sorted or how it determines.

If no:
1. Please move both inside of test_data. Then during the test `preCheck` setup you can move them to the root or to `.trunk/configs`.

linters/buildifier/buildifier.test.ts
TylerJang27: Inside of `.trunk/configs` should be fine, if that's what you're asking. If you meant something else, we can talk more offline

linters/buildifier/buildifier.test.ts
```diff
@@ -1,5 +1,32 @@
...
+};
 
 // TODO(Tyler): We will eventually need to add a couple more test cases involving failure modes.
 linterCheckTest({ linterName: "buildifier" });
-linterFmtTest({ linterName: "buildifier" });
+customLinterFmtTest({
+  linterName: "buildifier",
+  args: "-a",
+  testName: "add_tables",
+  pathsToSnapshot: [path.join(TEST_DATA, "add_tables.in.BUILD")],
```
TylerJang27, line 30: Can you append `path.join(TEST_DATA, "basic.in.bzl")` as well and also audit `buildifier_v4.0.1_basic.fmt.shot`? Looks like we lost some coverage here.

Because of the way we use `jest-specific-snapshot`, unfortunately it doesn't auto-detect the fact that we lost coverage.

linters/buildifier/plugin.yaml
```diff
@@ -23,21 +23,24 @@ lint:
       download: buildifier
       commands:
         - name: fix
-          run: buildifier --lint=fix --warnings=all "${target}"
```
TylerJang27, line 26: What's the rationale for removing `--warnings=all` here?

TylerJang27 (commented):
I appreciate the detail of the description. I suspect it will more likely than not be useful in the future (hopefully not near future).

There's a few pieces I don't fully understand about the direct_configs and you're testing strategy here. It's probably easiest to take this offline, then we can get the tests passing and this landed.

sxlijin:
Pushed up currently broken changes. Snapshots are missing; to repro the failures I was seeing with the framework, just run:

```
rm linters/buildifier/test_data/buildifier_v4.0.1_*fmt.shot
npm test linters/buildifier/
npm test linters/buildifier/  # this is the invocation that will fail
```

TylerJang27 (approved):
LGTM, fixed test to be explicit about input paths, and also added comment to plugin.yaml

TylerJang27:
Will add to linterFmtTest

PR #253, opened 2023-04-17
Add tool definitions to plugins

laurit17:
Adds a bunch of our trivially portable tools, and extends our existing testing framework to verify their installation.

Need to follow up with making sure runtimes are enabled for tools so that `asciinema` works - for now the error message is clear enough.


.github/filters.yaml
```diff
@@ -10,13 +10,25 @@ all-linters:
   - .trunk/trunk.yaml
   - package.json
 
+all-tools:
+  - tests/driver/**
```
TylerJang27, line 14: should we make this all except for lint_driver, and the reverse for `all-linters:`

tests/driver/driver.ts
```diff
@@ -0,0 +1,296 @@
...
+   * Run a specified trunk command with `args` and additional options.
+   * @param args arguments to run, excluding `trunk`
+   * @param execOptions
+   */
+  async runTrunk(
+    args: string[],
+    execOptions?: ExecOptions
+  ): Promise<{ stdout: string; stderr: string }> {
+    const trunkPath = ARGS.cliPath ?? "trunk";
+    return await execFilePromise(trunkPath, args.filter(Boolean), {
```
TylerJang27, line 282: nit: potentially not ok to use Boolean here, javascript does weird things imo, e.g. if you pass in an input "0" or "false" as part of your array

tests/driver/lint_driver.ts
```diff
@@ -0,0 +1,304 @@
...
+const getDebugger = (linter?: string) => {
+  if (!linter) {
+    // If a linter is not provided, provide a counter for easy distinction
+    return baseDebug.extend(`test${testNum++}`);
+  }
+  const numLinterTests = linterTests.get(linter);
+  const newNum = (numLinterTests ?? 0) + 1;
+  linterTests.set(linter, newNum);
+  return baseDebug.extend(linter).extend(`${newNum}`);
+};
```
TylerJang27, line 25: Really, this should be its own util class or something. Up to you on if you want to do it now or later

tests/driver/tool_driver.ts
```diff
@@ -0,0 +1,140 @@
...
+const getDebugger = (tool?: string) => {
+  if (!tool) {
+    // If a tool is not provided, provide a counter for easy distinction
+    return baseDebug.extend(`test${testNum++}`);
+  }
+  const numToolTests = toolTests.get(tool);
+  const newNum = (numToolTests ?? 0) + 1;
+  toolTests.set(tool, newNum);
+  return baseDebug.extend(tool).extend(`${newNum}`);
+};
```
TylerJang27, line 23: same

tests/driver/tool_driver.ts
```diff
@@ -0,0 +1,140 @@
...
+export interface TrunkToolRunResult {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+  /** Error thrown if the trunk invocation returned a nonzero exit code */
+  error?: Error;
+}
+
+export class TrunkToolDriver extends GenericTrunkDriver {
+  /** The name of the linter. If defined, enable the linter during setup. */
```
TylerJang27, line 37: nit: tool, change comments below too

tests/driver/tool_driver.ts
```diff
@@ -0,0 +1,140 @@
...
+  tool?: string;
+  /** Dictated version to enable based on logic of parsing environment variables. May be a version string or `LinterVersion` */
+  toEnableVersion?: string;
```
TylerJang27, line 40: Is your intention here to reuse `PLUGINS_TEST_LINTER_VERSION`? Might be worth a discussion

tests/driver/tool_driver.ts
```diff
@@ -0,0 +1,140 @@
...
+    super(testDir, setupSettings, getDebugger(tool));
+    this.tool = tool;
+    this.toEnableVersion = version;
+  }
+
+  /**
+   * Setup a sandbox test directory by copying in test contents and conditionally:
+   * 1. Creating a git repo
+   * 2. Dumping a newly generated trunk.yaml
+   * 3. Enabling the specified 'tool'
```
TylerJang27, line 54: 4. Sync it to make sure it's installed

tests/index.ts
```diff
@@ -47,7 +48,7 @@ const conditionalTest = (
   timeout?: number | undefined
 ) => (skipTest ? it.skip(name, fn, timeout) : it(name, fn, timeout));
 
-export type TestCallback = (driver: TrunkDriver) => unknown;
+export type TestCallback = (driver: TrunkLintDriver) => unknown;
```
TylerJang27, line 51: I think it might be useful to have preCheck for more intensive tools tests in the future. We should sync on long-term testing planning

tests/index.ts
```diff
@@ -119,6 +120,69 @@ export const setupDriver = (
...
+  afterEach(() => {
+    registerVersion(driver.enabledVersion);
+  });
+  return driver;
+};
+
+// Tool tests:
+// should take in: tool name
+// dir name
+// input/output
```
TylerJang27, line 153: nit: remove?

tests/index.ts
```diff
@@ -119,6 +120,69 @@ export const setupDriver = (
...
+  testConfigs: ToolTestConfig[];
+}) => {
+  describe(toolName, () => {
+    const driver = setupTrunkToolDriver(dirName, {}, toolName, toolVersion);
+    testConfigs.forEach(({ command, expectedOut, expectedErr, expectedExitCode }) => {
+      it(`should run "${command.join(" ")}" and exit with code ${expectedExitCode}`, async () => {
+        const { stdout, stderr, exitCode } = await driver.runTool(command);
+        expect(stdout).toContain(expectedOut);
+        expect(stderr).toContain(expectedErr);
+        expect(exitCode).toEqual(expectedExitCode);
```
TylerJang27, line 180: I would recommend a custom matcher here, since it can give more information at once in the object. Rn, if this fails, expect only tells you the first failure. There's objectContains as well, which might also be a fit, although it doesn't support substrings out of the box. Can be tabled for later if you want

tools/bazel/bazel.test.ts
```diff
@@ -0,0 +1,14 @@
+import { toolTest } from "tests";
+
+toolTest({
+  toolName: "bazel", // toolName
+  toolVersion: "6.0.0", // version
```
TylerJang27, line 5: Can we remove these comments, on all the tests?

.github/workflows/pr.yaml
```diff
@@ -39,6 +39,9 @@ jobs:
       all-linters: ${{ steps.post-filter.outputs.out }}
       # shortened paths to linter subdirs
       linters-files: ${{ steps.post-filter-paths.outputs.out }}
```
TylerJang27, line 41: Can we add nightly tests too? Can be its own job, doesn't need to be integrated with the rest rn, but I'd like to start having some viz on this

tools/gh/gh.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+toolTest({
+  toolName: "gh", // toolName
+  toolVersion: "2.27.0", // version
+  testConfigs: [
+    {
+      command: ["gh", "--version"],
+      expectedExitCode: 0,
+      expectedOut: "gh version 2.27.0",
+      expectedErr: "",
```
TylerJang27, line 11: Should we default this to an empty string? Same with exit code to 0?

TylerJang27 (commented):
Looks great! A few notes and maybe discussion points. Overall I wanted to highlight that we should sync on long term planning for testing here, but some standout notes:
1. tools should have a preCheck hook
2. tools should have the ability to run differently on different versions (see `skipTestIf`)
3. tools should maybe have some nicer defaults and the ability to run a series of conditions (I see you have `testConfigs` as a list here, which is a good start. Maybe we want elements to also be able to be callbacks, so that we can do setup between steps (e.g. `(driver)=>{driver.writeFile("a", "b");}`))
4. tool tests should be able to run on specific OSs (same as 2)

These don't need to be done now, but are probs necessary in some capacity in order to integrate with LUV.

tests/driver/driver.ts
laurit17: replaced

tools/helm/helm.test.ts
```diff
@@ -1,14 +1,12 @@
...
+  toolVersion: "3.9.4",
   testConfigs: [
-    {
+    makeToolTestConfig({
       command: ["helm", "version"],
-      expectedExitCode: 0,
-      expectedOut: `version.BuildInfo{Version:"v3.9.4"`,
-      expectedErr: "",
-    },
+      expectedOut: 'version.BuildInfo{Version:"v3.9.4"',
```
TylerJang27, line 9: nit: potentially worth a comment, potentially not. This is substring matching, correct. Does it usually include other info that's not worth having a "}"?

TylerJang27 (commented):
Important! Please remove `registerVersion(driver.enabledVersion);` from the tool driver setup code. This will start sending it to LUV, which we're not ready for yet (collisions could be a real problem), and we likely want to establish separate endpoints or codepaths for tools vs linters.

Once that is done, will stamp and seems good to land

TylerJang27 (approved):
LGTM!

PR #261, opened 2023-04-21
Initial commit of sourcery

fsargent:
https://docs.sourcery.ai/Welcome/


fsargent:
Still needs snapshots.
--fix does not work, as it modifies files in place rather than to stdout for `output: rewrite` to work.

TylerJang27:
> Still needs snapshots. --fix does not work, as it modifies files in place rather than to stdout for `output: rewrite` to work.

We supply the `in_place: true` option for cases like this, which should be sufficient. You can add it as a field to fix subcommand.

TylerJang27:
Getting this setup properly is proving to be a bit more difficult than initially expected. I'll have an additional PR up shortly for you to take a look at that should have some fixes to make this work with our CI and for standard use.

TylerJang27:
Hi @fsargent! Because getting sourcery working required a bit more documentation and CI setup, I've gone ahead and made this [PR](https://github.com/fsargent/plugins/pull/1). You can merge that into your branch, and that should be enough to get this ready to land!

TylerJang27:
Sorry, looks like I missed something. I'll investigate and get back to you soon

TylerJang27:
Ah, GitHub doesn't allow the secret modification that I've done here from a non-repository contributor. @fsargent, mind if I take over getting this landed via #264?

fsargent:
Sure

PR #262, opened 2023-04-22
Use native sarif support in brakeman

det:


TylerJang27 (approved):
LGTM. Interesting that we didn't already have this, since the KGV is 5.4.0. If brakeman sees an uptick in usage, we may want to backfill an older snapshot, but should be fine for now

PR #265, opened 2023-04-25
Update readme

dapirian:


.trunk/trunk.yaml
```diff
@@ -20,7 +21,7 @@ lint:
   ignore:
     - linters: [ALL]
       paths:
-        - "**/test_data" # required for golangci-lint, which runs on directories
```
TylerJang27, line 23: nit: I think this is useful to leave in on account of it looking odd that we require two seemingly identical globs

readme.md
```diff
@@ -1,40 +1,37 @@
...
-  </a>
-</p>
+<!-- markdownlint-disable first-line-heading -->
+
+[![Trunk.io](https://static.trunk.io/assets/trunk_plugins_logo.png)](https://trunk.io)
+
+[![docs](https://img.shields.io/badge/-docs-darkgreen?logo=readthedocs&logoColor=ffffff)][docs]
+[![slack](https://img.shields.io/badge/-slack-611f69?logo=slack)][slack]
+[![vscode](https://img.shields.io/visual-studio-marketplace/i/trunk.io?color=0078d7&label=vscode&logo=visualstudiocode)][vscode]
+[![openssf](https://api.securityscorecards.dev/projects/github.com/trunk-io/trunk-action/badge)](https://api.securityscorecards.dev/projects/github.com/trunk-io/trunk-action)
```
TylerJang27, line 8: This is the wrong badge link. Should be plugins, not action

readme.md
```diff
@@ -171,12 +168,18 @@ details.
 
 <br/>
 
+### Supported Trunk Actions
+
+You can think of Trunk Actions as IFTTT for your repository. An action is a command that is run in
+reaction to a specified trigger. Triggers can be git-hooks, file modifications, time-based, of
```
TylerJang27, line 174:
```suggestion
reaction to a specified trigger. Triggers can be git-hooks, file modifications, time-based, or
```

TylerJang27 (commented):
Please fix link, then ready to stamp

readme.md
dapirian: Wow nice catch, I never would have noticed that 🥇

.trunk/trunk.yaml
dapirian: ya thing is I thought this was actually for multiple linters and was more nuanced than the comment implies. Also we need to eliminate the need for this

TylerJang27: I see. IIRC it's just for golangci-lint at this point, since it's the only one with directory targets that we have turned on in this repo. It's not the easiest to eliminate the need for this, but I agree we should

TylerJang27 (approved):

.trunk/trunk.yaml
dapirian: fair enough - will re-add

PR #267, opened 2023-04-27
fix istioctl download declaration

laurit17:
does not have arm64 build for mac

### DIFF
--- a/tools/istioctl/plugin.yaml
+++ b/tools/istioctl/plugin.yaml
@@ -3,13 +3,13 @@ tools:
   downloads:
     - name: istioctl
       downloads:
-        - os:
-            linux: linux
-            macos: osx
+        - os: macos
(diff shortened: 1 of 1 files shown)


tools/istioctl/plugin.yaml
```diff
@@ -3,13 +3,13 @@ tools:
   downloads:
     - name: istioctl
       downloads:
-        - os:
-            linux: linux
-            macos: osx
+        - os: macos
+          url: https://github.com/istio/istio/releases/download/${version}/istioctl-${version}-osx-amd64.tar.gz
```
TylerJang27, line 7: do we need to specify a cpu in this section in order for it to be applied correctly?

laurit17: in that case we get the missing platform exception - working on that separately

TylerJang27 (approved):
lgtm

PR #275, opened 2023-05-02
Add http_proxy passthroughs for runtimes, move to `:-` default env var syntax

dapirian:


TylerJang27:
This breaks checkov locally. Investigating why

dapirian:
Update: we'll have to support this in the CLI as opposed to in config. The idea is to pass through these 2 env vars for all subcommands the Trunk cli/daemon runs

PR #277, opened 2023-05-04
Partial terragrunt integration

laurit17:
- add .envrc to plugins
- terraform tool integration
- integrating terragrunt fmt

I would also like to integrate terragrunt's validation tool - as mentioned in standup, it requires terraform to be in your `PATH`. However, even with `terraform` in path, I cannot get it to work properly. There is another environment variable, `TERRAGRUNT_TFPATH`, which allows me to get this to work, but we'd have to evaluate that to point to the `$(which terraform)`. So, punting on that integration for now.


linters/terragrunt/plugin.yaml
```diff
@@ -0,0 +1,55 @@
...
+          cpu:
+            x86_64: amd64
+            arm_64: arm64
+          url: https://github.com/gruntwork-io/terragrunt/releases/download/v${version}/terragrunt_${os}_${cpu}
+        - os: windows
+          cpu: x86_64
+          url: https://github.com/gruntwork-io/terragrunt/releases/download/v${version}/terragrunt_windows_amd64.exe
+  definitions:
+    - name: terragrunt
+      download: terragrunt
```
TylerJang27, line 39: Can you add `suggest_if: never`?

linters/terragrunt/plugin.yaml
```diff
@@ -0,0 +1,55 @@
...
+      files: [hcl]
+      environment:
+        - name: PATH
+          list: ["${linter}"]
+      commands:
+        - name: format
+          output: rewrite
+          run: terragrunt hclfmt ${target}
+          success_codes: [0]
+          in_place: true
```
TylerJang27, line 50: Can you add `sandbox_type: copy_targets`

tools/terraform/plugin.yaml
```diff
@@ -0,0 +1,22 @@
+version: 0.1
+tools:
+  downloads:
+    - name: terraform
```
TylerJang27, line 4: Is this going to collide with the terraform linter download? A bit worried about cli internals here

linters/terragrunt/terragrunt.test.ts
```diff
@@ -0,0 +1,3 @@
+import { linterFmtTest } from "tests";
+
+linterFmtTest({ linterName: "terragrunt" });
```
TylerJang27, line 3: So only the validation subcommand requires terraform to be present? And so we're fine for this subcommand?

TylerJang27 (commented):
Some comments that I'd like to have followup discussion, otherwise good

linters/terragrunt/terragrunt.test.ts
laurit17: yep, this one doesn't need terraform

tools/terraform/plugin.yaml
laurit17: fair enough - we won't have a collision, but this should be resolved somehow anyway

PR #278, opened 2023-05-05
Fixes windows downloads and parser/run stems

TylerJang27:
Adds several fixes to support Windows including:
1. Fixing some download URLs
2. Adding runtime stems to some parsers and actions
3. Adding platform-specific commands for known linters (e.g. pragma-once, google-java-format)
4. Adds supported_platforms restrictions to ruby and ansible-lint
5. Bumps CLI version in order to include some MacOS fixes and platform/supported platform support

Remaining todos:
- Align file patterns and matching to respect / and \\ (circleci config, actionlint)
- Audit remaining linters


linters/detekt/plugin.yaml
```diff
@@ -14,8 +14,9 @@ lint:
       commands:
         - name: lint
           output: sarif
+          # TODO(Tyler): This needs to be platform-dependent
           run:
-            detekt-cli --build-upon-default-config --config .detekt.yaml --input ${target,} --report
+            detekt-cli.bat --build-upon-default-config --config .detekt.yaml --input ${target,} --report
```
det, line 19: We already search for .exe in the CLI, we need to extend it to also look for bat and then we can revert this.

TylerJang27:
Test times are creeping up. Might be worth investigating soon

det (approved):

PR #283, opened 2023-05-12
WIP: workflow to call trunk-io/trunk-action/upgrade

sxlijin:


.github/workflows/upgrade_trunk.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # For trunk to create PRs
+      pull-requests: write # For trunk to create PRs
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      # >>> Install your own deps here (npm install, etc) <<<
+      - name: Trunk Upgrade
+        uses: trunk-io/trunk-action/upgrade@main
```
TylerJang27, line 19: We should do with args `--bleeding-edge` here

.github/workflows/upgrade_trunk.yaml
```diff
@@ -0,0 +1,19 @@
...
+  trunk_upgrade:
+    name: Upgrade Trunk
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # For trunk to create PRs
+      pull-requests: write # For trunk to create PRs
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      # >>> Install your own deps here (npm install, etc) <<<
```
TylerJang27, line 17: We don't have a `.trunk/setup-ci/action.yaml`, so we'll need to at least npm install here

PR #291, opened 2023-05-23
Add support for multiple languages for terrascan

laurit17:
note: will upgrade to 1.10.1 

fixes: https://linear.app/trunk/issue/TRUNK-6192/terraform-linter-integrations

### DIFF
--- a/linters/terrascan/plugin.yaml
+++ b/linters/terrascan/plugin.yaml
@@ -21,11 +21,13 @@ lint:
         - name: lint
           output: sarif
           is_security: true
-          run: terrascan scan --iac-file ${target} --output sarif
+          run: terrascan scan -i terraform --iac-file ${target} --output sarif
           read_output_from: stdout
-          files: [terraform] # terrascan supports - arm, cft, docker, helm, k8s, kustomize, terraform. However, we only support terraform and docker as filetypes.
+          # terrascan supports - arm, cft, docker, helm, k8s, kustomize, terraform. However, we only support terraform and docker
+          # as filetypes as it's not straightforward to map the other filetypes to a specific language.
+          files: [terraform]
           success_codes: [0, 3, 4, 5]
-          parser:
+          parser: # necessary to convert file paths from a strange custom format to a standard format
             runtime: python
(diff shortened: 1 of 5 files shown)


linters/terrascan/plugin.yaml
```diff
@@ -16,15 +16,25 @@ lint:
...
       known_good_version: 1.18.1
-      files: [terraform]
       suggest_if: config_present
       commands:
         - name: lint
           output: sarif
           is_security: true
           run: terrascan scan --iac-file ${target} --output sarif
+          read_output_from: stdout
+          files: [terraform] # terrascan supports - arm, cft, docker, helm, k8s, kustomize, terraform. However, we only support terraform and docker as filetypes.
```
TylerJang27, line 26: I presume the yaml-key detection for these other file types is nontrivial?

linters/terrascan/plugin.yaml
```diff
@@ -16,15 +16,25 @@ lint:
...
           success_codes: [0, 3, 4, 5]
+          parser:
+            runtime: python
+            run: ${plugin}/linters/terrascan/sarif_to_sarif.py
```
TylerJang27, line 30: nit: I remember we added this before, but might be nice to add a comment as to why we do this. IIRC, it was because the output was too verbose for a normal diagnostic

linters/terrascan/plugin.yaml
TylerJang27, line 26: does this run command not need a `-i terraform` or anything?

TylerJang27 (approved):
Looks good. A few other comments for sanity checks

linters/terrascan/plugin.yaml
laurit17: that's the default, but specified terraform to be more explicit.

linters/terrascan/plugin.yaml
laurit17: done

linters/terrascan/plugin.yaml
laurit17: yes - left a comment

PR #292, opened 2023-05-24
add python-interface as a file type

Ryang20718:
Add .pyi file extension as a file type. Contents should python interfaces/stubs.

Could also add under "python" since contents should just be python, but not sure if that's kosher.


linters/plugin.yaml
sxlijin, line 1: Thanks! We should also be able to update `linters/mypy/plugin.yaml` and add `python-interface` to the list of supported filetypes there.

Ryang20718: Added!

laurit17 (commented):
make sure you also add `comments: -hash` like for python.

Ryang20718:
Added. @laurit17 wondering why is the `comments: -hash` needed? Locally it worked without that?

TylerJang27:
> Added. @laurit17 wondering why is the `comments: -hash` needed? Locally it worked without that?

Hi @Ryang20718. It's necessary for some other functionality within trunk, specifically, supporting [trunk-ignore](https://docs.trunk.io/docs/ignoring-issues).

TylerJang27 (approved):

TylerJang27:
Thanks for the contribution! Note that you'll need to point your trunk.yaml plugin ref to `51e060ebfe5e19723265f83f715d95045a2916b4` in the short term until our next plugins release in order for your changes to take effect

Ryang20718:
Thanks Tyler for helping merge!

PR #294, opened 2023-05-26
feat(osv-scanner): make the diagnostics more useful 

sxlijin:
Eventually we should figure out how to surface the vulnerability details
as well, but for now, the error messages should at the minimum be useful
for users who don't know what they're looking at.

Setting details_url, prefixing the message, and setting the line number
should do the trick.


.trunk/trunk.yaml
```diff
@@ -15,6 +15,8 @@ plugins:
 
 lint:
   # enabled list inherited from plugin 'configs'
+  enabled:
+    - osv-scanner@1.3.3
```
puzzler7, line 19: sanity check: do we want osv-scanner to be enabled in the plugins repo?

puzzler7 (approved):

puzzler7:
Also the tests need to be updated

.trunk/trunk.yaml
sxlijin: yeah

PR #298, opened 2023-05-31
Add txtpbfmt

pat-trunk-io:


pat-trunk-io:
Current dependencies on/for this PR:
* main
  * **PR #298** <a href="https://app.graphite.dev/github/pr/trunk-io/plugins/298" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/trunk-io/plugins/298?utm_source=stack-comment).

pat-trunk-io:
@TylerJang27 
I got it to work with trunk fmt, but the jest tests don't seem to work.
The output in the snapshot is the same as the source file

pat-trunk-io:
> @TylerJang27 I got it to work with trunk fmt, but the jest tests don't seem to work. The output in the snapshot is the same as the source file

Nevermind, figured it out.
Ready for review

linters/txtpbfmt/plugin.yaml
```diff
@@ -0,0 +1,17 @@
...
+      commands:
+        - name: format
+          output: rewrite
+          run: txtpbfmt ${target}
+          success_codes: [0]
+          in_place: true
+          batch: true
+          formatter: true
+      suggest_if: files_present
+      known_good_version: 0.0.0-20230412060525-fa9f017c0ded
```
TylerJang27, line 17: Do you know what this version is supposed to represent/how stable is it?

Additionally, there are some cases in which we don't support semver pre-release tags other than `alpha`, `beta`, and `rc`. I'll need to play around with this a bit to make sure it doesn't break anything in the `trunk upgrade` flow.

linters/txtpbfmt/txtpbfmt.test.ts
```diff
@@ -0,0 +1,3 @@
+import { linterFmtTest } from "tests";
+
+linterFmtTest({ linterName: "txtpbfmt", namedTestPrefixes: ["test0", "test1"] });
```
TylerJang27, line 3: nit: namedTestPrefixes should be unnecessary here, but fine to leave it

TylerJang27 (approved):
Looks good! See a couple comments, I'm particularly worried about the semver issue, although I'll investigate more tomorrow. If possible, hold off on merging til then

linters/txtpbfmt/plugin.yaml
pat-trunk-io: They don't have tags sadly

TylerJang27:
@pat-trunk-io Just did a quick audit. This version tag should be fine. If we need to manually update people, we can bump the `known_good_version` and people will correctly pick it up during `trunk upgrade`. And since their repo doesn't have tagged releases, we don't have to worry about that logic. The one thing that this will not support is versioned subcommands, so that's something we'll need to keep in mind if it becomes relevant.

Going to try and fix the other plugins tests so this can land.

PR #299, opened 2023-05-31
Enable trufflehog by default instead of gitleaks

det:
Tested various repos and trufflehog appears to be a superset of what gitleaks catches, with the exception of gitleak's `generic-api-key`. I have to imagine trufflehog doesn't catch this by default intentionally.


TylerJang27 (approved):
LGTM. This should be able to land immediately. No need to wait on my other fixes

PR #300, opened 2023-06-05
Tell trufflehog not to update itself

det:
I got this when running trufflehog with `trunk -a --filter=trufflehog`:
```
        {"level":"error","ts":"2023-06-05T13:06:31-07:00","logger":"trufflehog","msg":"error occured with trufflehog updater 🐷","error":"cannot move binary (exec: \"mv\": executable file not found in $PATH)"}
```

Apparently, trufflehog will try to auto-update itself if we don't pass `--no-update`.


puzzler7 (approved):
LGTM

PR #306, opened 2023-06-08
Enable gt in tools

pat-trunk-io:


TylerJang27 (approved):
Looks good to me. Gonna ask Sam about the check failure

TylerJang27:
> Looks good to me. Gonna ask Sam about the check failure

This can land since it's not a required check. Following up with another PR to get that services Check passing.

PR #311, opened 2023-06-12
Some messaging in truck-check-pre-push-always

siddharthab:
Remind the user that trunk is running in non-interactive mode, and if there are
issues, then the user can try running the command manually from a terminal.


TylerJang27 (approved):
LGTM. Thanks for following up on this. I'll go ahead and handle any upkeep to help get this landed!

TylerJang27:
Pointing out one other thing we've added recently but haven't documented yet, you can add the following to your trunk.yaml to get more detailed information about what trunk actions are running when they run:

```yaml
actions:
  output_level: <hidden/short/verbose>
```

PR #312, opened 2023-06-13
Create NPM install pre-push hook as a guard before running tests

laurit17:
Here's how it looks:

Screenshot 2023-06-12 at 8 26 02 PM

fixes: https://linear.app/trunk/issue/TRUNK-7252/git-hooks-in-plugins-fail-if-jest-not-installed


package-lock.json
```diff
@@ -21,7 +21,7 @@
         "@typescript-eslint/parser": "^5.50.0",
         "caller": "^1.1.0",
         "debug": "^4.3.4",
-        "eslint": "8.39.0",
+        "eslint": "^8.30.0",
```
TylerJang27, line 24: Can you quickly sanity check that this still runs on a typescript file in the repo? There was an eslint bug that caused me to pint his beforehand.

TylerJang27 (approved):
Looks good, and this should help a lot. See sanity check

PR #313, opened 2023-06-13
Add fuzzy test matching for non-idempotent linters

TylerJang27:
Adds a fuzzy check test that operates as follows:
- Provide most of the same arguments as other tests
- The test strips the snapshot of its `issues` so that it's only an empty list
- Create an `expected_issues.json` file with the expected issues
- Call `fuzzyLinterCheckTest` and pass in a fuzzyMatcher, which includes the expected issues and a minimum required threshold for overlap

The current overlap for osv-scanner is 30 of 51, and the current overlap for trivy is 12 of 16. Using the custom matcher, it still prints all of the output issues, and it still does the normal assertion behavior for the rest of the landing state. If the test fails due to a significant drift in fileIssues, we can either adjust the threshold or modify `expected_issues.json` to include a more updated list. I estimate that upkeep for these non-idempotent linters will be reduced to once very few months, barring other breaking changes.

This implementation leaves room for version-specific expected issues still, allowing us to preserve the existing snapshot logic as needed. Regardless, this shouldn't be necessary in most cases since these linters tend to only have one versioned snapshot.


tests/index.ts
```diff
@@ -12,17 +12,29 @@ import {
...
 // support this. Adding this patch/override allows us to call it with custom matchers, as in the case of `landingStateWrapper`.
+type OwnMatcher<Params extends unknown[]> = (
+  this: jest.MatcherContext,
+  actual: unknown,
+  ...params: Params
+) => jest.CustomMatcherResult;
+
 declare global {
-  // trunk-ignore(eslint/@typescript-eslint/no-namespace)
+  // eslint-disable-next-line @typescript-eslint/no-namespace
```
TylerJang27, line 30: I prefer doing the eslint ignore (rather than trunk) directly here since the extension's warning is quite glaring.

laurit17 (approved):
LGTM. Does the absolute number of issues emitted vary at all? You might consider fixing this to a percentage (or even a low absolute number -- maybe just 5 or 10, given we're unlikely to have a bug where we're only parsing some but not all diagnostics). Doesn't seem obviously better to me though.

TylerJang27:
> LGTM. Does the absolute number of issues emitted vary at all? You might consider fixing this to a percentage (or even a low absolute number -- maybe just 5 or 10, given we're unlikely to have a bug where we're only parsing some but not all diagnostics). Doesn't seem obviously better to me though.

The absolute number of issues tends to remain relatively static. Percentage is a good suggestion, but I think it's fine for now. Higher number feels right for now but it might need some tuning as we go.

PR #314, opened 2023-06-13
Make sure all tools have known_good_version

laurit17:
Also correct one where the version is not a semver


TylerJang27 (approved):
Good catch. Thanks for this. Any inclination on how we'd want to prioritize validated upgrades for these?

laurit17:
> Good catch. Thanks for this. Any inclination on how we'd want to prioritize validated upgrades for these?

it's a fairly high priority, probably in the next sprint or two. For now, we want to make sure `trunk tools enable <unversioned_tool>` doesn't cause any surprises.

PR #315, opened 2023-06-14
fix: added checkov to security linters

puzzler7:
Also alphabetically sorted the `Security` row in the readme.


TylerJang27 (approved):
LGTM. Note that this will start failing release tests until the next release is out since we've changed the definition and no longer tag with release.

To get tests passing, you might have missed a snapshot or 2 to update. You can prepend `PLUGINS_TEST_LINTER_VERSION=...` to ensure the test runs with a specific snapshot (or generates one if not found)

PR #319, opened 2023-06-16
Bump cli version

TylerJang27:
Bump the cli version. We want people to be using the daemon fixes, so this and higher should be the standard for this repo for now.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -2,7 +2,7 @@ version: 0.1
 
 # version used for local trunk runs and testing
 cli:
-  version: 1.10.1-beta.35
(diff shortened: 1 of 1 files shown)


puzzler7 (approved):

det (approved):

.trunk/trunk.yaml
```diff
@@ -2,7 +2,7 @@ version: 0.1
 
 # version used for local trunk runs and testing
 cli:
-  version: 1.10.1-beta.35
+  version: 1.10.1-beta.39
```
det, line 5:
```suggestion
  version: 1.10.1-beta.40
```

det (approved):

PR #326, opened 2023-06-22
Don't autosuggest nancy

TylerJang27:
Nancy can have some painfully flaky installs, even with our retry threshold. Let's not auto-recommend it anymore for go repos.


TylerJang27:
@det. Discussed this with Sam and Mav. It causes some decently common install failures, so we don't want to autosuggest it. Not sure if this merits another follow-up release...

puzzler7 (approved):

PR #338, opened 2023-06-28
Remove file type osv-lockfiles

TylerJang27:
Since we merge lint ignores, this can consolidate with the rest of our lockfiles. Follow-up to PR and thread from #208.


sxlijin (approved):
can we add go.sum to the osv test data to assert that it's excluded?

puzzler7 (approved):
LGTM, +1 on Sam's comment

PR #342, opened 2023-06-29
Perltidy and Perlcritic

dex4er:
Hi, there are new linters.

Because there is no good source of Perl binaries and usually Perl is installed OS-natively and most of tools can be installed locally then I didn't add `perl` engine nor separate versioned tools.

### DIFF
--- a/.github/actions/linter_tests/action.yaml
+++ b/.github/actions/linter_tests/action.yaml
@@ -38,6 +38,10 @@ runs:
       shell: bash
       working-directory: ${{ inputs.path }}
 
+    - name: Install packages
+      run: sudo apt-get -y install libperl-critic-perl perltidy
+      shell: bash
+
     - name: Install dependencies
       run: npm ci
       shell: bash
--- a/.github/actions/linter_tests/action.yaml
+++ b/.github/actions/linter_tests/action.yaml
@@ -39,7 +39,17 @@ runs:
       working-directory: ${{ inputs.path }}
 
     - name: Install packages
-      run: sudo apt-get -y install libperl-critic-perl perltidy
+      run: |
+        case "$RUNNER_OS" in
+          Linux)
+            sudo apt-get -y install libperl-critic-perl perltidy
+            ;;
+          macOS)
+            brew install cpm
+            cpm install -g --no-test --color Perl::Critic Perl::Tidy
+            brew unlink perl && brew link perl
+            ;;
+        esac
       shell: bash
 
     - name: Install dependencies
--- a/.github/actions/tool_tests/action.yaml
+++ b/.github/actions/tool_tests/action.yaml
@@ -31,6 +31,10 @@ runs:
       shell: bash
       working-directory: ${{ inputs.path }}
 
+    - name: Install packages
+      run: sudo apt-get -y install libperl-critic-perl perltidy
+      shell: bash
+
     - name: Install dependencies
       run: npm ci
       shell: bash
--- a/.github/actions/tool_tests/action.yaml
+++ b/.github/actions/tool_tests/action.yaml
@@ -31,10 +31,6 @@ runs:
       shell: bash
       working-directory: ${{ inputs.path }}
 
-    - name: Install packages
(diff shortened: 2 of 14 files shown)


TylerJang27:
Thanks for the PR! The linter definitions looks great, and we're excited to have some new perl support. I'm going to do some investigation on my end to see if there is a way we can get hermetic installs set up. I'll loop back in the next day or two with some feedback and my findings

.github/actions/tool_tests/action.yaml
```diff
@@ -31,6 +31,10 @@ runs:
       shell: bash
       working-directory: ${{ inputs.path }}
 
+    - name: Install packages
+      run: sudo apt-get -y install libperl-critic-perl perltidy
+      shell: bash
```
TylerJang27, line 36: Go ahead and remove this. Tool tests are separate, and since this is only integrating linters/formatters, it's not needed here

.github/actions/linter_tests/action.yaml
```diff
@@ -38,6 +38,10 @@ runs:
       shell: bash
       working-directory: ${{ inputs.path }}
 
+    - name: Install packages
```
TylerJang27, line 41:
```suggestion
    - name: Install packages
      if: env.RUNNER_OS == 'Linux'
```

linters/perltidy/perltidy.test.ts
```diff
@@ -0,0 +1,3 @@
+import { linterFmtTest } from "tests";
+
+linterFmtTest({ linterName: "perltidy", namedTestPrefixes: ["basic"] });
```
TylerJang27, line 3:
```suggestion
import { linterFmtTest } from "tests";
import { skipOS }  from "tests/utils";

linterFmtTest({ linterName: "perltidy", namedTestPrefixes: ["basic"], skipTestIf: skipOS(["darwin"]) });
```

linters/perlcritic/perlcritic.test.ts
```diff
@@ -0,0 +1,3 @@
+import { linterCheckTest } from "tests";
+
+linterCheckTest({ linterName: "perlcritic", namedTestPrefixes: ["basic"] });
```
TylerJang27, line 3:
```suggestion
import { linterFmtTest } from "tests";
import { skipOS }  from "tests/utils";

linterCheckTest({ linterName: "perltidy", namedTestPrefixes: ["basic"], skipTestIf: skipOS(["darwin"]) });
```

TylerJang27 (commented):
Thanks for the PR. I've made a few suggestions in order to get tests passing, but otherwise this should be good to land as is!

.github/actions/tool_tests/action.yaml
dex4er: Done.

.github/actions/linter_tests/action.yaml
dex4er: latest-macos supports Perl so maybe it is better to add it. I use MacOS M1 as my main platform so I don't want to disable tests for this OS.

linters/perltidy/perltidy.test.ts
dex4er: Not skipped if macOS is configured in GHA workflow.

linters/perlcritic/perlcritic.test.ts
dex4er: Not skipped if macOS is configured in GHA workflow.

dex4er:
@TylerJang27 Please confirm we can leave the Perl for macOS on GitHub Actions.

.github/actions/linter_tests/action.yaml
TylerJang27: Hi thanks for addressing the comments. We can run this test on Mac, it was just missing the correct installation step. Assuming this next CI run passes I'll do a final pass and stamp

TylerJang27 (approved):
LGTM. Thanks for the PR!

PR #347, opened 2023-06-30
GraphQL schema linter integration

laurit17:
Establishing the pattern of using a tool dependency


TylerJang27 (approved):
LGTM, couple things:

1. Will want to give the linter a pass myself when I get back before next plugins release; we've landed a couple linters lately that could've used a couple more config options.
2. Looks like the test filter is not being applied correctly. Either add the suggested flag (with prepending `--`) to `append-args` or add an `if` check on the action step.

PR #349, opened 2023-07-01
Forward line information from trufflehog output

det:


det:
@TylerJang27  Is out output somehow non-deterministic? Looks like it re-ordered some entries? Also is the `// trunk-upgrade-validation:RELEASE` line important?

puzzler7 (approved):
LGTM pending Tyler's approval

TylerJang27 (commented):
> @ Tyler Is out output somehow non-deterministic? Looks like it re-ordered some entries? Also is the // trunk-upgrade-validation:RELEASE line important?

LGTM. Please remove the `// trunk-upgrade-validation:RELEASE` from the snapshot. That's only supposed to be there if the linter definition is stable but the snapshot was updated since our last release. By removing it, LUV will use the snapshot tagged at v0.0.21 rather than manually pulling this version in, allowing LUV to continue working as intended until the next release.

PR #350, opened 2023-07-02
tflint v0.47.0

dex4er:
tflint plugin didn't work with tflint 0.47.0. I think it is because this version shows:

```
Command line arguments support was dropped in v0.47. Use --chdir or --filter instead.
```

My projects work well with the configuration from this PR.

### DIFF
--- a/linters/tflint/plugin.yaml
+++ b/linters/tflint/plugin.yaml
@@ -28,6 +28,15 @@ lint:
           read_output_from: stdout
           run_linter_from: root_directory
           run_from_root_target: .tflint.hcl
+          version: ">=0.47.0"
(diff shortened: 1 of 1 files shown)


TylerJang27:
Hi @dex4er! Thanks for identifying this issue and for the PR. Unfortunately our automated testing for tflint is constrained by rate limits, otherwise it would have caught this. I've gone ahead and made an alternative PR #351 with a slightly different approach that I think more closely mirrors the intended tflint invocation. If you could checkout that branch and test it locally to make sure it fits your needs, I'd appreciate it.

dex4er:
I'm closing it in favour of PR #351

PR #352, opened 2023-07-06
Use a python parser for rubocop

det:


linters/rubocop/rubocop_to_sarif.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+import json
+import sys
+
+
+def map_severity(severity):
+    if severity in ["convetion", "refactor", "info"]:
```
gejustin, line 8: Looks like there might be a typo here:

```suggestion
    if severity in ["convention", "refactor", "info"]:
```

det: Thanks!

TylerJang27 (approved):
LGTM

PR #354, opened 2023-07-07
Update graphql-schema-linter to only act on `.graphqls` files

dapirian:


laurit17 (commented):
so I had started a branch on this but meant to follow up - does this make us too restrictive? a lot of people use graphql extension for their schema files it seems

dapirian:
> so I had started a branch on this but meant to follow up - does this make us too restrictive? a lot of people use graphql extension for their schema files it seems

well there's usually many graphql files and fewer graphql schema files, and graphql-schema-linter fails on non-schema files with `graphql-schema-linter/invalid-graphql-schema`. We could get fancy and try to infer whether a graphql file is schema or not, but I think easier in the short term is to require `graphqls` as the suffix, which is a semi-popular convention for graphql schema but not ubiquitous

laurit17 (approved):

PR #359, opened 2023-07-11
Forward ssh auth sock in LFS git hooks

det:
Otherwise users will have to type their ssh keyfile passwords out even after running `ssh-add`


EliSchleifer (approved):
Lgtm

PR #360, opened 2023-07-13
Wrong Rome Filename

brunowego:
Hi folks! Currently the name of filename config is wrong. The correct name is `rome.json`. Look https://github.com/rome/tools/blob/main/rome.json.


dapirian (approved):
Nice catch!

PR #372, opened 2023-07-18
Convert sourcery to a trunk tool

laurit17:
did this one separately because I don't have a sourcery account and was too lazy to make one so I could test locally


TylerJang27 (approved):
LGTM. Talked offline about SOURCERY_TOKEN and its merits/drawbacks

PR #375, opened 2023-07-19
Add tool download URLs for Windows

det:


TylerJang27 (approved):
LGTM, also see [tyler/trunk-8003-make-tools-play-nice-on-windows for](https://github.com/trunk-io/trunk/pull/9929) for ease of testability with `trunk tools install`

linters/nancy/plugin.yaml
```diff
@@ -40,4 +40,8 @@ lint:
           list: ["${runtime}", "${linter}", "${env.PATH}"]
         - name: USER
           value: ${env.USER}
+          optional: true
+        - name: USERNAME
```
TylerJang27, line 44: I think we want this flipped? It requires USER but USERNAME might be set in the env

PR #384, opened 2023-07-22
Fix osv-scanner line numbers

dapirian:
The previous logic was reading lines of a lockfile line-by-line inside the loop that iterated issues, but as the issue iteration marched on, the read position in the lockfile did too (didn't ever reset via fseek(0)), so what ended up happening was some issues would have line numbers but most wouldn't. This changes it to just read the entire lockfile into a string, and then iterate the lines of that string multiple times.


linters/osv-scanner/osv_to_sarif.py
```diff
@@ -86,17 +87,19 @@ def main(argv):
...
+
                 description = (
                     f'Vulnerability in {pkg["name"]}@{pkg["version"]}: {message}'
                 )
 
-                lines = [
-                    zero_indexed + 1
-                    for zero_indexed, line in lockfile_lines
-                    if pkg["name"] in line and pkg["version"] in line
-                ]
```
dapirian, line 97: note: this list comprehension logic is more lines, more confusing, and less optimal, than just writing it as a regular for loop 😉

det (approved):

PR #396, opened 2023-07-26
Fix all tests on linux arm64

pat-trunk-io:
This changes make all tests pass on Linux arm64, with the exception of the following tests because they don't have arm64 builds

Tools:
   - gt
 
 Linters:
   - Scalafmt 
   - oxipng
   - sourcery
   
   Notable change: Kube-lint only has x86 binaries, but can be build with go install, so switched it to that.


linters/kube-linter/plugin.yaml
```diff
@@ -28,5 +13,5 @@ lint:
       environment:
         - name: PATH
           list: ["${linter}"]
-      known_good_version: 0.5.0
+      known_good_version: 0.6.4
```
TylerJang27, line 16: Please rename snapshots too. There should always be a snapshot for the `known_good_version` value (this wouldn't cause tests to fail, but it's a repo cleanliness thing and makes it easier to keep track of what is running what)

linters/scalafmt/scalafmt.test.ts
```diff
@@ -1,24 +1,32 @@
...
   const contents = `version = ${driver.enabledVersion ?? "3.4.3"}\nrunner.dialect = scala3`;
   driver.writeFile(".scalafmt.conf", contents);
 };
 
+const skipTestIfLambda = (version?: string | undefined) => {
+  const os_version = skipOS(["win32"])(version);
+  if (os_version) {
+    return os_version;
+  }
+  return skipCPUOS(["linux"], ["arm64"])(version);
```
TylerJang27, line 16: nit: Can we just `||` this?

linters/semgrep/plugin.yaml
```diff
@@ -27,7 +27,7 @@ lint:
       environment:
         - name: PATH
           list: ["${env.PATH}"]
-      known_good_version: 1.7.0
+      known_good_version: 1.33.2
```
TylerJang27, line 30: Same, plz rename snapshot

linters/taplo/plugin.yaml
```diff
@@ -97,7 +97,7 @@ lint:
       environment:
         - name: PATH
           list: ["${linter}"]
-      known_good_version: release-cli-0.6.0
```
TylerJang27, line 100: Snapshot as well for both in this file plz

tests/utils/index.ts
```diff
@@ -225,6 +225,20 @@ export const getVersionsForTest = (
...
+ */
+export const skipCPU = (excludedCPU: string[]) => (_version?: string) =>
+  excludedCPU.length === 0 || excludedCPU.includes(process.arch);
+
+/**
+ * Helper callback that skips a test if the OS and CPU arch is included in excludedCPU and excludedOS.
+ * Intended to be passed to `skipTestIf`.
+ */
+export const skipCPUOS = (excludedOS: string[], excludedCPU: string[]) => (_version?: string) =>
+  excludedCPU.includes(process.arch) && excludedOS.includes(process.platform);
```
TylerJang27, line 240: I think it would be better if this were pairwise or struct'd somehow. E.g.:

`skipCPUOS([{os: "linux", cpu: "aarch64"}]);`

Not a strong opinion, but I think it would allow us to hit contrasting cases more easily and cleanly

TylerJang27 (approved):
Couple comments, some requiring snapshot renaming. Otherwise good

PR #397, opened 2023-07-27
Fix Nightly linter cli integration tests

pat-trunk-io:
I think it's only missing an apt update in the nightly linter/cli integration test


TylerJang27 (approved):
LGTM. Can you also add:
1. Have you tested running this manually off your branch?
2. `continue_on_error: true` to this step?

PR #399, opened 2023-07-28
Add variant of trufflehog that scans git history

det:
- Runs once on the repo root instead of individual files.
- Tells trufflehog what the upstream is.
- Trufflehog scans all commits in between.
- Ensure that timestamps are deterministic so that commit hashes are deterministic.


TylerJang27 (commented):
Could we get some quick test coverage of this? Probably something similar to the `git-diff-check` test setup?

det:
> Could we get some quick test coverage of this? Probably something similar to the `git-diff-check` test setup?

Thanks for the pointer, done.

linters/trufflehog/trufflehog.test.ts
```diff
@@ -1,2 +1,28 @@
...
+import { TrunkLintDriver } from "tests/driver";
+
 linterCheckTest({ linterName: "trufflehog" });
+
+// Create secrets that exist both only in history, and also in the working tree.
+const preCheck = async (driver: TrunkLintDriver) => {
+  if (driver.gitDriver) {
+    await driver.gitDriver.checkoutLocalBranch("branch-with-secrets");
+    driver.copyFileFromRootTo("linters/trufflehog/test_data/secrets.in.py", "secrets.in.py");
+    driver.copyFileFromRootTo("linters/trufflehog/test_data/secrets.in.py", "secrets2.in.py");
```
TylerJang27, line 11: At least the first call shouldn't be necessary? Everything in the test_data folder gets copied into the sandbox

TylerJang27 (approved):
LGTM. If the commit SHA still isn't playing nice in the tests, you can modify the `normalizeMessage` call to look for `commit <sha>` and replace `<sha>`, although we might lose a little bit of coverage

linters/trufflehog/trufflehog.test.ts
det: I think `customLinterCheckTest` doesn't copy those files? The `git add` fails without copying it myself first.

det:
> LGTM. If the commit SHA still isn't playing nice in the tests, you can modify the `normalizeMessage` call to look for `commit <sha>` and replace `<sha>`, although we might lose a little bit of coverage

Determinism was more effort than it was worth. Added the ability to normalize the landing state after running linters to `customLinterCheckTest`. I'll wait for your sign off before landing.

linters/trufflehog/trufflehog.test.ts
TylerJang27:
```
    await driver.setUp();
    if (preCheck) {
      // preCheck is not always async, but we must await in case it is.
      await preCheck(driver);
      driver.debug("Finished running custom preCheck hook");
    }
```
preCheck should be called after any initial linter driver setup, so it should be there? Potentially worth investigating but priority is getting this landed

linters/trufflehog/trufflehog.test.ts
```diff
@@ -1,2 +1,34 @@
...
+  }
+};
+
+// Rewrite the landing state to remove non-deterministic commit hashes.
+const normalizeLandingState = (landingState: LandingState) => {
+  if (landingState.issues) {
+    for (const issue of landingState.issues) {
+      if (issue.message) {
+        issue.message = issue.message.replace(/commit [0-9a-f]{40}/g, "commit <hash>");
+      }
```
TylerJang27, line 25: Mild preference for this in `landing_state.ts` rather than as an additional input to `customLinterCheckTest`, but this should be fine and not a strong opinion.

TylerJang27 (approved):

PR #400, opened 2023-07-28
Fix rel path resolution

pat-trunk-io:


tests/utils/index.ts
```diff
@@ -37,6 +37,7 @@ const normalizePath = (value?: string): string | undefined => {
       return value;
     }
     path.resolve(REPO_ROOT, value);
+    return value;
```
TylerJang27, line 40: `return path.resolve(REPO_ROOT, value);`?

TylerJang27 (commented):
This is definitely a bug. I think returning value is not correct though?

TylerJang27 (approved):

PR #404, opened 2023-07-31
Fix autopep8 by pinning pycodestyle version

TylerJang27:
`pycodestyle` [released](https://pypi.org/project/pycodestyle/#history) a breaking version on Saturday that breaks all new installs of autopep8. We can pin an older version in order to temporarily resolve this for our users. I know we said we want to avoid this sort of a fix, but I don't think we have a fix for now other than pinning.


TylerJang27:
Related: https://github.com/hhatto/autopep8/issues/689

det (approved):

PR #410, opened 2023-08-01
Fix prisma and dustilock for Windows

TylerJang27:
- Fix prisma by finding the correct binary
- Fix dustilock by expanding the regex


puzzler7 (approved):
Dustilock changes look good - I don't have context on the prisma issues

PR #412, opened 2023-08-01
Reconvert linters to use tools

laurit17:
fixes: https://linear.app/trunk/issue/TRUNK-6640/have-linter-definitions-use-tools


linters/clang-format/plugin.yaml
```diff
@@ -1,26 +1,35 @@
...
-            x86_64: x86_64
-            arm_64: arm64
-          url: https://trunk.io/releases/clang-format-${version}-${os}-${cpu}.tar.gz
-        # TODO(chris): Windows download
+      download: clang-format
+      shims: [clang-format]
+      known_good_version: 14.0.1
+      environment:
+        - name: PATH
+          list: ["${linter}/bin"]
```
TylerJang27, line 31: Do we support `${tool}` here or does it implicitly resolve from linter?

linters/renovate/plugin.yaml
```diff
@@ -1,4 +1,11 @@
 version: 0.1
+tools:
+  definitions:
```
TylerJang27, line 3: Can this linter not be refactored to use the tool?

linters/taplo/plugin.yaml
```diff
@@ -97,7 +102,7 @@ lint:
       environment:
         - name: PATH
           list: ["${linter}"]
-      known_good_version: 0.8.0
+      known_good_version: release-cli-0.6.0
```
TylerJang27, line 105: I think this should be 0.8.0. Was this downgrade intentional?

linters/terragrunt/plugin.yaml
```diff
@@ -21,22 +21,9 @@ tools:
...
-            x86_64: amd64
-            arm_64: arm64
-          url: https://github.com/gruntwork-io/terragrunt/releases/download/v${version}/terragrunt_${os}_${cpu}
-        - os: windows
-          cpu: x86_64
-          url: https://github.com/gruntwork-io/terragrunt/releases/download/v${version}/terragrunt_windows_amd64.exe
   definitions:
     - name: terragrunt
-      download: terragrunt
+      tools: [terragrunt]
```
TylerJang27, line 26: Are we not including terraform here yet?

TylerJang27 (approved):
Looks good except for a couple sanity checks. Also:
1. Is the plan moving forward that any new linters should be defined as tools first if they are able to?
2. `prisma` I think may want its tools and linters combined

TylerJang27:
I think the testing failures are arising from the (accidental?) taplo KGV bump

linters/renovate/plugin.yaml
laurit17: no, it can't - due to the symlinks. can follow up to support that functionality with tools.

linters/terragrunt/plugin.yaml
laurit17: only needed for validate (if we integrate it)

linters/taplo/plugin.yaml
laurit17: ah that was a mistake - merging and all that. fixed, thanks for the catch

PR #416, opened 2023-08-02
Integrate tools into trunk tools

laurit17:
there are...a lot of these

### DIFF
--- a/tools/act/plugin.yaml
+++ b/tools/act/plugin.yaml
@@ -1,0 +1,22 @@
+version: 0.1
+downloads:
+  - name: act
+    downloads:
+      - os:
+          linux: Linux
+          macos: Darwin
+        cpu:
+          x86_64: x86_64
+          arm_64: arm64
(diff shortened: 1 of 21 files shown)


tools/act/plugin.yaml
```diff
@@ -0,0 +1,22 @@
...
+          linux: Linux
+          macos: Darwin
+        cpu:
+          x86_64: x86_64
+          arm_64: arm64
+        url: https://github.com/nektos/act/releases/download/v${version}/act_${os}_${cpu}.tar.gz
+      - os:
+          windows: Windows
+        cpu:
+          arm_64: arm64
```
TylerJang27, line 15: They don't have an x86_64 download?

TylerJang27 (approved):
LGTM

tools/act/plugin.yaml
laurit17: not for that version - added a version specifier

PR #421, opened 2023-08-03
Add a pre-commit hook to block secrets using trufflehog

det:
It would be nice if we had a way to run this in the same trunk invocation as `trunk fmt`, but we don't support that at the moment.


TylerJang27 (approved):
LGTM. Please add to the top-level README.md as well

PR #423, opened 2023-08-07
Add shims to runtimes

laurit17:
pretty straightforward - this is everything in the bin/ directory. Not sure if we need everything here, but can't hurt.


det (approved):
lgtm, we should consider adding glob support to shims.

PR #430, opened 2023-08-09
Add remaining pre-commit-hooks

pat-trunk-io:


det (approved):

det:
Looks like you need to define the xml file type.

linters/pre-commit-hooks/plugin.yaml
```diff
@@ -19,3 +19,148 @@ lint:
           success_codes: [0, 1]
           files:
             - ALL
+        - name: check-added-large-files
+          enabled: false
+          output: pass_fail
+          read_output_from: stdout
+          run: check-added-large-files ${target}
```
TylerJang27, line 26: Can we add `python3` to the front of these if appropriate? Shebangs don't work for these executions on Windows.

.github/workflows/pr.yaml
```diff
@@ -115,18 +115,50 @@ jobs:
...
   # Run tests against all linters for known_good_version and latest version
-  linter_tests:
-    name: Linter Tests ${{ matrix.os }}
-    runs-on: [self-hosted, "${{ matrix.os }}"]
+  linter_tests_linux:
+    name: Linter Tests Linux
+    runs-on: public-amd64-2xlarge
+    needs: detect_changes
+    if:
+      needs.detect_changes.outputs.linters == 'true' || needs.detect_changes.outputs.all-linters ==
```
TylerJang27, line 123: If we end up keeping this, can we merge it back into the matrix and add a new field to the object for the correct runner name?

linters/pre-commit-hooks/plugin.yaml
pat-trunk-io: Done

TylerJang27 (approved):
LGTM

linters/pre-commit-hooks/plugin.yaml
TylerJang27: False alarm. This was previously WAI, the files are .exe on Windows and inside of `Scripts/`, so no need for the python3 bit.

pat-trunk-io: Updated

PR #431, opened 2023-08-10
Separate functional eponymous tools into their own directories

TylerJang27:
The following tools were moved because they constitute their own semantic "tools" outside of just being linters/formatters:
- circleci (also added a snapshot for its latest release)
- renovate
- sourcery
- terraform

The following tools were left for a later date:
- ansible/ansible-lint (upgrade concerns)
- buf (non-eponymous naming concerns)

Also made some small testing and generating improvements

### DIFF
--- a/linters/renovate/plugin.yaml
+++ b/linters/renovate/plugin.yaml
@@ -1,11 +1,4 @@
 version: 0.1
-tools:
-  definitions:
-    - name: renovate
-      runtime: node
-      package: renovate
-      shims: [renovate]
-      known_good_version: 34.122.0
 lint:
   files:
     - name: renovate-config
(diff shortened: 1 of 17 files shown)


linters/renovate/plugin.yaml
```diff
@@ -17,8 +10,7 @@ lint:
   definitions:
     - name: renovate
       files: [renovate-config]
-      runtime: node
-      package: renovate
+      tools: [renovate]
```
TylerJang27, line 13: Any reason this wasn't done in the first pass?

laurit17: probably that's just an oversight

laurit17 (approved):

PR #440, opened 2023-08-17
feat: add trivy secret scanning

puzzler7:
Adds trivy secret scanning - the reason this wasn't showing up before was because trivy by default doesn't scan for secrets in files with `/test` in the path. Also reworks the trivy tests to have separate tests for each subcommand.

```
secrets.py:2:0
  2:0  high  AWS Access Key ID       trivy/aws-access-key-id    
  3:0  high  AWS Secret Access Key   trivy/aws-secret-access-key
 11:0  high  Asymmetric Private Key  trivy/private-key          
```


linters/trivy/config_expected_issues.json
```diff
@@ -0,0 +1,158 @@
+[
```
TylerJang27, line 1: Please also delete the `expected_issues.json` file

linters/trivy/plugin.yaml
```diff
@@ -26,23 +26,40 @@ tools:
...
           read_output_from: stdout
           # Trivy does not support batching
           batch: false
           is_security: true
           parser:
             runtime: python
-            run: python3 ${plugin}/linters/trivy/trivy_fs_to_sarif.py
+            run: python3 ${plugin}/linters/trivy/trivy_fs_vuln_to_sarif.py
+        - name: fs-secret
+          files: [ALL]
```
TylerJang27, line 48: Before we end up turning this on for everyone, couple verification questions:
1. How does performance compare to something like trufflehog?
2. How does secret scanning effectiveness compare to something like trufflehog?
3. What's the relative noisiness for our monorepo?

If any of these answers are in the negative direction, we may want to consider disabling this subcommand by default.

linters/trivy/trivy.test.ts
```diff
@@ -1,16 +1,65 @@
...
 import { FileIssue } from "tests/types";
+import { TEST_DATA } from "tests/utils";
 import { createFuzzyMatcher } from "tests/utils/landing_state";
 
-// trunk-ignore(eslint/@typescript-eslint/no-unsafe-assignment)
-const expectedFileIssues = JSON.parse(
-  fs.readFileSync(path.resolve(__dirname, "expected_issues.json")).toString()
-);
+const callbackGenerator =
+  (command: string, otherPreCheck?: (driver: TrunkLintDriver) => void): TestCallback =>
```
TylerJang27, line 10: nit: can `otherPreCheck?` be a `TestCallback` type?

linters/trivy/trivy.test.ts
```diff
@@ -1,16 +1,65 @@
...
+    );
+    driver.writeFile(trunkYamlPath, newContents);
+    if (otherPreCheck) {
+      otherPreCheck(driver);
+    }
+  };
+
+const configExpectedFileIssues = JSON.parse(
+  fs.readFileSync(path.resolve(__dirname, "config_expected_issues.json")).toString(),
+) as FileIssue[];
```
TylerJang27, line 27: nit: I would actually prefer this as a magic number rather than parsing here, just in case if we mess it up accidentally, we'll be asserting something like a 0 issue match, which will always be true

linters/trivy/trivy.test.ts
```diff
@@ -1,16 +1,65 @@
...
+
+fuzzyLinterCheckTest({
+  linterName: "trivy",
+  testName: "config",
+  args: "-a -y",
```
TylerJang27, line 32: remove `-y`. No autofixes

TylerJang27 (commented):
Couple discussion points. Looks mostly good

linters/trivy/plugin.yaml
puzzler7: I think this should be disabled by default.

1. It's slower than trufflehog.
2. It's about the same, I think.
3. Almost none - it has the added benefit of default ignoring our many fake secrets in test files.

Despite these, it has a few benefits over trufflehog - you can define your own regexes to search for and you can define regexes to not alert on (e.g. if you only use AWS keys, you can only search for those, making this tool faster). I think having trufflehog enabled for everybody, and then trivy secret scanning available but disabled by default for users looking to customize their secret scanning further is the best way to go.

TylerJang27 (approved):
LGTM

PR #468, opened 2023-09-07
Add dotnet format

TylerJang27:
Add [dotnet-format](https://github.com/dotnet/format/tree/main), which is run from [dotnet](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format). Some downloads are provided via [Microsoft](https://dotnet.microsoft.com/en-us/download/dotnet/7.0), but not in an easy templatizable format.

This formatter does not work on Mac M1 for unknown reasons. The first run it appears to _sometimes_ spawn a zombie process that we can't terminate from, causing Trunk to hang. I've disabled the M1 download for that reason until we can figure out a fix.

This will require a minor version bump since we bump the required trunk version.

As a side note, dotnet-format sometimes leaves around other file artifacts as a result of its run, but I believe this is expected behavior within the dotnet ecosystem and these should appropriately be git-ignored.


EliSchleifer:
@TylerJang27  - is this waiting on anything now?

TylerJang27:
> @TylerJang27 - is this waiting on anything now?

This works on all platforms except for M1 Mac, where it hangs for unknown reasons (and only on first run). I've been doing an ongoing investigation whenever I have free cycles. I've relayed this to the original user who requested this as well (who uses an M1 Mac).

pat-trunk-io (approved):

PR #472, opened 2023-09-13
Set hadolint timeout

TylerJang27:
We should lower our global timeout as well, but 20s is a generous timeout we can use for hadolint to avoid the macos issues.


det:
Can we do 60s here? In a CI environment you may have a very overloaded machine, or do you think 20s is ok even there?

TylerJang27:
> Can we do 60s here? In a CI environment you may have a very overloaded machine, or do you think 20s is ok even there?

In practice hadolint was completing instantly for me, but 60 sounds reasonable

det (approved):

PR #467, opened 2023-09-07
Use new `run_from` syntax

TylerJang27:
Replaces `run_linter_from` and `run_from_root_target` and `run_from_root_file`. Bumps the required cli version

PR #473, opened 2023-09-13
Upgrade trunk

TylerJang27:
This uses the latest version with the MacOS fixes, so we should be able to use crashpad again.


TylerJang27:
Attaching this as part of #467

PR #481, opened 2023-09-19
Allow actionlint to run shellcheck on run blocks

det:
actionlint will run shellcheck on run blocks as long as it is in path.


laurit17 (approved):
woo!

TylerJang27:
Blocked by https://linear.app/trunk/issue/TRUNK-8916/on-upgrade-should-enable-tools-that-are-missing-versions

EliSchleifer:
@det  -  Is this landable now? Looks like blockers have landed

det:
> @det - Is this landable now? Looks like blockers have landed

Needs 1.17.0 which is releasing now.

PR #485, opened 2023-09-21
Update jq definition to handle release label including jq-

EliSchleifer:
JQ was not properly defined with semver parsing from GH release data. Fixed now. And removed the version check


TylerJang27 (approved):
Might run into some issues like iwyu if we build out more infra for the testing here, but this looks good for current setup

PR #486, opened 2023-09-22
osv-scanner wants to know where GO is located so it can scan it

EliSchleifer:
osv-scanner wants to be able to scan the go version for security issues. Need to pass that along now for scanning. 

This makes a dep on go in definition for the linter so we pull in go. Really we'd want to add go to path if it exists on the system.

### DIFF
--- a/linters/osv-scanner/plugin.yaml
+++ b/linters/osv-scanner/plugin.yaml
@@ -45,7 +45,7 @@ lint:
       suggest_if: files_present
       environment:
         - name: PATH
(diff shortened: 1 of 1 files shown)


TylerJang27:
@EliSchleifer can we close this? I believe it was addressed by https://github.com/trunk-io/plugins/pull/544 (we discussed briefly as a team and decided using system go would be a smoother experience for most users)

EliSchleifer:
yup - closed.

PR #487, opened 2023-09-24
Update oxipng to use new prompt support

EliSchleifer:
- Update oxipng to use new prompt support
- Move minimum required version


EliSchleifer:
Not sure what is happening here to break cspell....weird failure. Maybe you can provide insight @TylerJang27

TylerJang27 (approved):
Looks good. cspell appears to be acting up as of last Friday and only on Linux. Will do some digging. If this is urgent it can merge if that's the only failure, but otherwise I will try to get up a fix first thing Monday

EliSchleifer:
Cool. Not urgent - can wait till monday

PR #498, opened 2023-10-03
Consider pyproject.toml for bandit

siddharthab:
Checked that both .bandit and pyproject.toml are respected, with preference given to .bandit.

### DIFF
--- a/linters/bandit/plugin.yaml
+++ b/linters/bandit/plugin.yaml
@@ -12,12 +12,12 @@ lint:
       files: [python]
       tools: [bandit]
       suggest_if: files_present
-      direct_configs: [.bandit]
+      direct_configs: [.bandit, pyproject.toml]
       commands:
         - name: lint
           # Custom parser type defined in the trunk cli to handle bandit's JSON output.
           output: bandit
-          run: bandit --exit-zero --ini .bandit --format json --output ${tmpfile} ${target}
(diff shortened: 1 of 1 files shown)


siddharthab:
Can I add the flag conditionally in the run command based on whether the file exists? If not, I can replace the command with a script.

TylerJang27:
We don't have any way to add flags conditionally based on if config files exist, although it is a common case that we are aware of. If bandit doesn't error when a `-c` arg doesn't exist, you can add it, but otherwise it will run into issues. I will take a closer look at this tomorrow to see what our options are

TylerJang27:
Unfortunately bandit fails if the config file doesn't exist, so we can't support this run command change OOTB, and we'd prefer not to layer this with an additional script. Instead, users will have to override this execution in their trunk.yaml if they want bandit to check their pyproject.toml. I've created a feature request you can upvote to handle this in the future: https://features.trunk.io/check/p/support-configuration-files-in-run-commands-if-they-exist

This does highlight a bug, however, that we are not respecting `pyproject.toml` as an `affects_cache` element. I will put up a separate PR that fixes this.

siddharthab:
> This does highlight a bug, however, that we are not respecting `pyproject.toml` as an `affects_cache` element.

bandit does not read `pyproject.toml` if not explicitly specified on the command line, so it does not affect the cache.

PR #501, opened 2023-10-04
Add platformio

pat-trunk-io:


laurit17 (changes requested):
can you add a health check please? Then you will only need `toolInstallTest`. thanks

pat-trunk-io:
> can you add a health check please? Then you will only need `toolInstallTest`. thanks

Done

laurit17 (approved):
thanks!!

PR #505, opened 2023-10-05
fix(trufflehog): linter name referenced by --filter

error9900:
I am unable to commit to my repo after enabling the `trufflehog-pre-commit` action.

`.trunk/logs/cli.log` is showing:

```
[10-04 20:50:00] [client.cc:19] GRPC Failed (error code INTERNAL): [--filter] 'trufflehog-git' is not enabled
```

It looks like the `--filter` option of the `trunk check` command in the actions definition is incorrectly referencing the `trufflehog` linter...

Running this in the CLI:
```
trunk check -t "git-commit" --upstream=HEAD --filter=trufflehog-git
```

Results in:
```
✖ [--filter] 'trufflehog-git' is not enabled
```

This runs fine in the CLI:
```
trunk check -t "git-commit" --upstream=HEAD --filter=trufflehog
```

### DIFF
--- a/linters/trufflehog/plugin.yaml
+++ b/linters/trufflehog/plugin.yaml
@@ -80,7 +80,7 @@ actions:
     - id: trufflehog-pre-commit
       description: Don't allow secrets in commits (via Trufflehog)
       display_name: Trufflehog Pre-Commit Hook
-      run: trunk check -t "git-commit" --upstream=HEAD --filter=trufflehog-git
+      run: trunk check -t "git-commit" --upstream=HEAD --filter=trufflehog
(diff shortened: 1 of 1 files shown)


TylerJang27:
Thanks for the PR and attention to detail! However, this action is actually correct as written. We intentionally use `trufflehog-git` for this action, since it is git-aware. As you noticed, we can improve on how we communicate this to our users. I'm going to close this PR, but I will add some additional comments and documentation to make this trufflehog distinction clearer.

error9900:
The clarification in #509 helps. Thanks!

PR #520, opened 2023-10-11
Skip sourcery linter tests when token not provided

TylerJang27:
TODO


TylerJang27:
Using https://github.com/trunk-io/plugins/pull/521 instead

PR #521, opened 2023-10-11
Skip sourcery linter tests

TylerJang27:
Prevent linter sourcery tests from running when the token is not set. This is designed to make local development simpler and allow for a better experience from [forks](https://github.com/trunk-io/plugins/pull/517). Adds a log if the test is skipped.

Tested on https://github.com/trunk-io/plugins/pull/520.


EliSchleifer (approved):
LGTM

PR #523, opened 2023-10-11
Add linter for xml

pat-trunk-io:


linters/prettier/plugin.yaml
```diff
@@ -28,6 +19,27 @@ lint:
...
+            - typescript
+            - yaml
+            - css
+            - sass
+            - html
+            - markdown
+            - json
+            - javascript
+            - graphql
+            - prettier_supported_configs
```
TylerJang27, line 32: Strictly speaking, this is correct. But just wanted to call out that we may see a couple reports when people upgrade because I know a handful of people overrode prettier filetypes in order to support some of the prettier plugins

linters/prettier/plugin.yaml
```diff
@@ -28,6 +19,27 @@ lint:
...
+        - name: format-xml
+          output: rewrite
+          run: prettier --plugin=@prettier/plugin-xml -w ${target}
+          success_codes: [0]
+          batch: true
+          in_place: true
+          formatter: true
+          version: ">=3.0.0"
```
TylerJang27, line 40: Calling out for posterity that the latest @prettier/plugin-xml only works on prettier >=3.0.0, and you have to manually downgrade in order to support older versions

TylerJang27 (approved):

PR #530, opened 2023-10-18
fix(trufflehog): improve signal-to-noise of trufflehog diagnostics

sxlijin:
We have three known issues with Trufflehog currently:

1. it sometimes does not find secrets correctly, because we pass it `file.foo\n` instead of `file.foo` in some cases, due to a bug internal to the CLI.
1. it suffers from a high false positive rate on a number of detectors, which lowers its signal-to-noise ratio.
1. it reproducibly reports the wrong line numbers for diagnostics at times.

This PR resolves the first two issues:

1. by collapsing the `trufflehog.lint` command onto a single line, we bypass the path bug (we'll have a fix out for this in the CLI soon as well)
1. we set `--only-verified` when running trufflehog to filter out false positives - although verification requires network access, it implements network access by testing API keys and the like against the service provider API in question, so there's no risk of sending compromised credentials to Trufflehog or any other 3p service (although if a detector wrongly tested a real API key against the wrong service, and said other service was breached, that would be a leak of the API key in question - but this is an acceptable tradeoff)

We're still following up on a resolution for incorrect line numbers.

---

I don't like how this causes trufflehog to stop reporting SSH private keys, but it's a tradeoff I'm willing to make.


TylerJang27 (approved):
LGTM, assuming we are all committed to the verification change. Some testing callouts:

You will need to run:
```
PLUGINS_TEST_LINTER_VERSION=Snapshots npm test trufflehog -- -u
PLUGINS_TEST_UPDATE_SNAPSHOTS=true npm test trufflehog # a new release from today fixed _a_ line number bug
```

TylerJang27:
Can you also set `cache_results: false`

TylerJang27:
Actually, we should use `cache_ttl` if we can

sxlijin:
Added `cache_ttl: 1h`, thanks for the suggestion.

PR #531, opened 2023-10-18
fix(trufflehog): report the correct line numbers

puzzler7:
Trufflehog can report wildly incorrect line numbers for secrets, and even report multiple incorrect issues for a given secret. This PR changes our trufflehog parser to scan the file for the secret and report the correct line number. As the trufflehog can report duplicates, we only output a lint issue once for a given secret.

This is tested with `wrong_line_number.in.ts`, which trufflehog reports 4 issues on incorrect lines for 2 identical secrets. The parser finds these two secrets on their separate lines.

This only fixes the issue for trufflehog filesystem - I'm not sure if this an issue in trufflehog git, and the files may be changed/deleted since the past.


puzzler7:
Moving to draft for now while we wait on other resolutions for this problem

sxlijin:
Since trufflehog hasn't addressed the underlying issue and does not seem to be treating it with the appropriate severity, I've asked mav to resume work on this.

linters/actionlint/test_data/actionlint_v1.6.26_CUSTOM.check.shot
```diff
@@ -1,5 +1,4 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
-// trunk-upgrade-validation:RELEASE
 
 exports[`Testing linter actionlint test CUSTOM 1`] = `
 {
```
puzzler7, line 4: This change (and similar) is not related to trufflehog, but something @Tyler asked me to do to update the tests.

det (approved):

linters/trufflehog/test_data/trufflehog_git_v3.40.0_CUSTOM.check.shot
sxlijin, line 1: do we know why we're losing coverage here?

linters/trufflehog/test_data/trufflehog_v3.31.3_secrets.check.shot
sxlijin, line 1: same coverage loss

linters/trufflehog/test_data/wrong_line_number.in.ts
sxlijin, line 1: let's use the repros shared in https://github.com/trufflesecurity/trufflehog/issues/1876:

* https://gist.githubusercontent.com/det/1526b4c16d0e07ac023d75c912a68658/raw/c3061c14a811205a65cbdcf0065bd3c11d88bfcb/test.txt
* https://github.com/trufflesecurity/trufflehog/issues/1876#issuecomment-1771721567

and also a test for ssh private key leaks, since those are multiline by definition

linters/trufflehog/trufflehog_to_sarif.py
```diff
@@ -48,11 +75,11 @@ def main(argv):
...
         if "Filesystem" in vuln_json["SourceMetadata"]["Data"]:
+            secret = vuln_json["Raw"]
             path = vuln_json["SourceMetadata"]["Data"]["Filesystem"]["file"]
-            line_number = (
-                int(vuln_json["SourceMetadata"]["Data"]["Filesystem"].get("line", "0"))
-                + 1
-            )
+            line_number = find_line_number(secret, path)
+            if line_number is None:
+                continue
```
sxlijin, line 82: ugh... i'm kind of torn on this behavior. i don't know if we want to fallback to the trufflehog-specified line number or line 0.

discarding the diagnostic feels wrong though- i don't think we've ever seen trufflehog just completely hallucinate an issue, it's always just wrong location, right?

linters/trufflehog/trufflehog_to_sarif.py
```diff
@@ -28,6 +28,33 @@ def to_result_sarif(path: str, line_number: int, vuln_id: str, description: str)
     }
 
 
+secret_line_cache = {}
```
sxlijin, line 31:
```suggestion
secret_lineno_cache = {}
```

linters/trufflehog/trufflehog_to_sarif.py
```diff
@@ -28,6 +28,33 @@ def to_result_sarif(path: str, line_number: int, vuln_id: str, description: str)
...
+        file_cache[path] = open(path).readlines()
+
+    if secret not in secret_line_cache:
+        secret_line_cache[secret] = []
+
+    secret_length = len(secret.splitlines())
+    lines = file_cache[path]
+
+    for lineno in range(1, len(lines) - secret_length):
+        check_window = lines[lineno : lineno + secret_length]
```
sxlijin, line 46: let's lift this out into a helper, `sliding_window`, c.f. itertools example https://docs.python.org/3/library/itertools.html:

```
def sliding_window(iterable, n):
    # sliding_window('ABCDEFG', 4) --> ABCD BCDE CDEF DEFG
    it = iter(iterable)
    window = collections.deque(islice(it, n-1), maxlen=n)
    for x in it:
        window.append(x)
        yield tuple(window)
```

sxlijin (commented):
looking good. concerned about why we're not surfacing ssh keys though- do we need more tests for those?

linters/trufflehog/trufflehog_to_sarif.py
puzzler7: Trufflehog can report the same issue multiple times - for instance, the test file I added reports 4 secrets, reporting each of the secrets twice.

linters/trufflehog/test_data/wrong_line_number.in.ts
puzzler7: I redacted the file similarly but I'd prefer to stick with this repro, as trufflehog reports each secret twice.

linters/trufflehog/test_data/trufflehog_git_v3.40.0_CUSTOM.check.shot
puzzler7: I think the coverage loss is because trufflehog is considering the ssh key as unverified, and as we're now running trufflehog in only-verified mode, we don't report it.

linters/trufflehog/trufflehog_to_sarif.py
sxlijin: gah, ffs

let's add this as an inline comment here as well (i know you mentioned it above)

sxlijin (commented):
do we know if there are any multiline secret formats that we can add to wrong_line_number.in.ts to verify that the window logic works as intended?

if there are none, let's do a manual test with an SSH key and removing `--only-verified` from `trufflehog/plugin.yaml`

puzzler7:
> if there are none, let's do a manual test with an SSH key
couldn't find any, and the verified secrets that trufflehog tests with get pulled from google cloud so I couldn't use those. However, removing the `--only-verified` and running trunk check does report the multiline private key on the correct line.

sxlijin (approved):

PR #532, opened 2023-10-19
fix: trufflehog 3.60.1 has a backwards-breaking line number change

sxlijin:
Since we're waiting on a meeting with Trufflehog to merge #531


linters/trufflehog/test_data/trufflehog_v3.40.0_secrets.check.shot
```diff
@@ -19,7 +18,7 @@ b3BlbnNzaC1rZXktdjEAAAAACmFl",
       "file": "test_data/secrets.in.py",
       "isSecurity": true,
       "level": "LEVEL_HIGH",
-      "line": "8",
+      "line": "7",
```
sxlijin, line 21: IDK what happened here- is the constraint syntax not correct?

linters/trufflehog/plugin.yaml
```diff
@@ -34,7 +34,23 @@ lint:
           cache_ttl: 1h
           parser:
             runtime: python
-            run: python3 ${plugin}/linters/trufflehog/trufflehog_to_sarif.py
+            run:
+              python3 ${plugin}/linters/trufflehog/trufflehog_to_sarif.py
+              --parse-mode=1-indexed-linenos
```
TylerJang27, line 39: optional: I would prefer just doing this exactly as we do it for `ruff` (passing in 1 or 0 for argv[1]), but this is alright

linters/trufflehog/plugin.yaml
```diff
@@ -65,7 +81,27 @@ lint:
           cache_results: false
           parser:
             runtime: python
-            run: python3 ${plugin}/linters/trufflehog/trufflehog_to_sarif.py
+            run:
+              python3 ${plugin}/linters/trufflehog/trufflehog_to_sarif.py
+              --parse-mode=1-indexed-linenos
+        - name: lint
+          output: sarif
+          version: <3.60.1
```
TylerJang27, line 89: same

linters/trufflehog/plugin.yaml
sxlijin: I hear what you're saying about overkill, and it was definitely in the back of my mind, but I think this makes the plugin.yaml much more readable so I'm willing to make that tradeoff.

TylerJang27: Fair enough. Not a hill I'd die on, but I stand by my opinion. Fine to land this parser line as is

TylerJang27 (approved):
LGTM. I don't know why the trufflehog-git snapshots didn't require updating. That feels off to me... But I will investigate in a follow-up

PR #534, opened 2023-10-23
Fix terrascan

TylerJang27:
Terrascan is failing tests on [MacOS nightly](https://github.com/trunk-io/plugins/actions/runs/6585061572/job/17890774904) and Windows but not on Linux. They haven't released an update in a while, so this is likely a slight regression due to our PATH changes. Nevertheless, this is the only linter that's having a problem, and since terrascan appears to need to be git aware, we simply will add the SYSTEM path here.

### DIFF
--- a/linters/terrascan/plugin.yaml
+++ b/linters/terrascan/plugin.yaml
@@ -51,5 +51,5 @@ lint:
         run: terrascan version
       environment:
         - name: PATH
-          list: ["${linter}"]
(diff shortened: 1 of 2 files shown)


det (approved):

TylerJang27:
Missed the mark on this. Even adding a prepare_run for `terrascan init` does not fix it (or using a tools install). Seems to be their issue, +1'd the [bug report](https://github.com/tenable/terrascan/issues/1629).

PR #535, opened 2023-10-24
Checkov decode uri

TylerJang27:
Fix checkov output for non-ascii filepaths, which previously were treated as discards. Tested locally with such files.


TylerJang27:
Waiting on the next CLI prod release to land this, since I don't want it to block the next plugins release necessarily

laurit17 (approved):

PR #539, opened 2023-10-26
Update README.md

error9900:
List `prettier` for additional supported languages.

Supported languages listed here: https://prettier.io/docs/en/


EliSchleifer (approved):
LGTM. Thanks

TylerJang27:
Do we want to list those in our README.md if we don't support those filetypes out of the box (all require plugins)?

EliSchleifer:
Any reason not to bring in those plugins by default?

TylerJang27:
I stand corrected. These added file types are all ones we already support. Prettier has some other plugins that support additional filetypes or frameworks, but these are all included in the base and ones that we already have config for.

PR #542, opened 2023-10-28
Run linux tests on self-hosted linux runners.

EliSchleifer:
We've spun up a small cluster of ubuntu machines that can run our public repo tests. We'll be adding some self hosted mac runners when they arrive - so we can speed up these flows


TylerJang27 (approved):
LGTM. Can we allowlist these runners for actionlint?

EliSchleifer:
> LGTM. Can we allowlist these runners for actionlint?

done

PR #544, opened 2023-10-31
Fix OSV scanner env

laurit17:
Needs to access system path for go sometimes. also, having ${linter} in the path is redundant if you have a tool.


TylerJang27 (commented):
Can we do this by adding `runtime: go` instead?

laurit17:
> Can/should we do this by adding `runtime: go` instead?

Good call - it's installable as a go pkg.

det (approved):

PR #552, opened 2023-11-08
Support sourcery open source license

TylerJang27:
The sandbox needs to have a symlink to the repo root .git in order to be able to work for the open source license. This is what we use and should unblock any users who use trunk+sourcery with sourcery's open source license.

As a result of the need to create this symlink, we also necessarily can't support Windows for the near term (anyone on Windows with a paid token can still run it with some minor overrides).


pat-trunk-io (approved):
lgtm

PR #556, opened 2023-11-09
Add fix command for standardrb

TylerJang27:
Adds the recommended fix command to make standardrb a formatter. We still retain the lint subcommand, which outputs issues in excess of what the formatter (or the unsafe formatter) can apply. This seems like intended behavior and a wholehearted improvement.

Did an audit of these changes with [fastlane](https://github.com/fastlane/fastlane) and [CocoaPods](https://github.com/CocoaPods/CocoaPods), both worked well and seem reasonable, with the exception one file that got stuck in a loop.


pat-trunk-io (approved):
lgtm

PR #564, opened 2023-11-20
Fix eslint test packages and setup

TylerJang27:
Our eslint execution isn't super LUV-faithful, but regardless this was starting to fail on older versions. This fix will unblock eslint snapshot version tests. Fixes errors seen [here](https://github.com/trunk-io/trunk/actions/runs/6927792518/job/18842336047).


pat-trunk-io (approved):
lgtm

PR #576, opened 2023-12-05
Make sourcery work on all platforms

TylerJang27:
Use additional template resolution for symlinks in order to better support the creation of sandboxes for the open source sourcery license.

This will require a CLI version bump in order to pass tests (this causes a linter runtime error on existing versions).

### DIFF
--- a/linters/sourcery/plugin.yaml
+++ b/linters/sourcery/plugin.yaml
@@ -5,20 +5,16 @@ lint:
       files: [python]
       commands:
         - name: lint
-          platforms: [linux, macos]
           output: regex
-          run: bash -c "ln -s ${REPO_DIR}/.git .git; sourcery review  --no-summary --csv ${target}"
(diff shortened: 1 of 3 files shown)


TylerJang27:
For some reason, while this is much cleaner on Linux+Mac, on Windows when executing in a sandbox, sourcery is not able to resolve the open source license. Normally on Linux/Mac, it reads the symlink (equivalent of doing `git remote -v`) and does some more logic there. But even though `git` works fine with this on Windows, `sourcery` doesn't recognize it as open source when it's a symlink. This is as far as I've gotten, and I'm tabling this investigation for now

TylerJang27:
I think this is out of scope for now. There are some issues with Windows symlink creation that make this difficult to achieve for the open source case within the context of the Trunk execution model. If additional users ask for better sourcery Windows support we can invest more time, but for now closing this.

PR #580, opened 2023-12-07
Add support for graphite `>=1.0.0`

det:


tools/gt/plugin.yaml
```diff
@@ -16,6 +16,14 @@ downloads:
         url: https://github.com/withgraphite/graphite-cli/releases/download/v${version}/gt-${os}.exe
 tools:
   definitions:
+    - name: gt
```
TylerJang27, line 19: As a sanity check would be nice to have another toolInstallTest for this

pat-trunk-io (approved):
lgtm

PR #592, opened 2023-12-15
Ensure trunk knows when checkov fails

det:
Python will exit 1 when printing a backtrace from a crash so we want trunk to know that is a failure.
Tell checkov to always exit 0, even when there are issues, so we can rely on that.


pat-trunk-io (approved):
lgtm

PR #593, opened 2023-12-18
Add command name to codespell definition

TylerJang27:
Not sure how this got missed the first time. If it comes up again, we'll want to add a repo test for it. This would also be a good use case for the "trunk-doctor" or other tool.


EliSchleifer (approved):
LGTM

PR #606, opened 2024-01-02
Fix run_from for iwyu

TylerJang27:
This was "working" before because the default is to just fallback to the workspace root.

Also includes better test coverage for this and the relevant CLI `run_from: ${compile_command}` fix.

Also fixes a prettier snapshot with the latest upstream testing changes.


TylerJang27:
Needs new CLI changes before the test coverage component can land

sxlijin (approved):

PR #610, opened 2024-01-04
Run linter tests against upstream

TylerJang27:
This change has been a long-time coming, but now it's more feasible on our faster runners. Runs (virtually) all of our linters on the upstream as well, which improves coverage, and requires a lot of snapshot updates.

This also normalizes using `TEST_DATA` rather than `"-a"` as a `customLinterCheckTest` input because `-a` has other effects on results. Otherwise, there's not really any additional overhead for devs to worry about.

_This is a big one but it's mostly snapshots_. The only tests with real substantive changes are the `clang-tidy` and `terrascan` tests.


pat-trunk-io (approved):
Very large change, but LGTM

PR #628, opened 2024-01-09
DEBUG: Investigate buf issues

TylerJang27:
TODO


TylerJang27:
Could not reproduce issue from https://github.com/trunk-io/plugins/actions/runs/7458262133/job/20291909004

PR #635, opened 2024-01-18
Add additional stylelint configs

TylerJang27:
Stylelint has expanded its set of [configs](https://github.com/trunk-io/plugins/pull/new/tyler/stylelint-improvements) since we added our integration. Updates the configs and their order of precedence to reflect this.


drwpow (approved):
I think that captures everything!

PR #636, opened 2024-01-18
Add .mjs extension for eslint and prettier

drwpow:
## Changes

Adds `.mjs` extension for ESLint and Prettier plugins (note: `.mjs` is always a valid option whenever any `.js` extension is accepted, and Node automatically treats it as the same, just in enforced ESM mode)

### DIFF
--- a/linters/eslint/plugin.yaml
+++ b/linters/eslint/plugin.yaml
@@ -23,6 +23,7 @@ lint:
         - .eslintrc.cjs
         - .eslintrc.js
         - .eslintrc.json
+        - .eslintrc.mjs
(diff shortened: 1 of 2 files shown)


TylerJang27 (approved):
Thanks for catching this... Does this mean we'll also need to update `cspell` and a bunch of our other node linters?

drwpow:
> Thanks for catching this... Does this mean we'll also need to update `cspell` and a bunch of our other node linters?

I haven’t used cspell personally but probably! Anywhere `.js` can be used, `.mjs` and `.cjs` should be supported as well. They’re official Node extensions.

PR #644, opened 2024-01-29
Add support for cmake-format@0.6.13

tmpsantos:
Add support for the most well known tool for formatting CMake files.

https://cmake-format.readthedocs.io/

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ trunk check enable {linter}
 | C#              | [dotnet-format]                                                                                                      |
 | CircleCI Config | [circleci]                                                                                                           |
 | Cloudformation  | [cfnlint], [checkov]                                                                                                 |
+| CMake           | [cmake-format]                                                                                                       |
 | CSS, SCSS       | [stylelint], [prettier]                                                                                              |
 | Cue             | [cue-fmt]                                                                                                            |
 | Docker          | [hadolint], [checkov]                                                                                                |
@@ -98,6 +99,7 @@ trunk check enable {linter}
 [clang-format]: https://clang.llvm.org/docs/ClangFormat.html
 [clang-tidy]: https://clang.llvm.org/extra/clang-tidy/
 [clippy]: https://github.com/rust-lang/rust-clippy#readme
+[cmake-format]: https://cmake-format.readthedocs.io/en/latest
 [codespell]: https://github.com/codespell-project/codespell#readme
 [cspell]: https://github.com/streetsidesoftware/cspell#readme
 [cue-fmt]: https://cuelang.org/
--- a/linters/cmake-format/cmake-format.test.ts
+++ b/linters/cmake-format/cmake-format.test.ts
@@ -1,0 +1,3 @@
+import { linterFmtTest } from "tests";
+
+linterFmtTest({ linterName: "cmake-format" });
--- a/linters/cmake-format/plugin.yaml
+++ b/linters/cmake-format/plugin.yaml
@@ -1,0 +1,32 @@
+version: 0.1
+tools:
+  definitions:
+    - name: cmake-format
+      runtime: python
+      package: cmake-format
+      shims: [cmake-format]
+      known_good_version: 0.6.13
+lint:
+  definitions:
+    - name: cmake-format
+      files: [cmake]
+      commands:
+        - name: fix
+          formatter: true
+          output: rewrite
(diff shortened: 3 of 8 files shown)


linters/cmake-format/plugin.yaml
```diff
@@ -0,0 +1,30 @@
...
+lint:
+  definitions:
+    - name: cmake-format
+      files: [cmake]
+      commands:
+        - name: lint
+          run: cmake-format --check ${target}
+          success_codes: [0, 1]
+          read_output_from: stderr
+          output: pass_fail
```
TylerJang27, line 18: What's the main value of having this lint subcommand here? Imo we can just include the `name: fix` subcommand and that should be sufficient. When I run `trunk check enable cmake-format; trunk check --force --filter=cmake-format linters/cmake-format`, I get `ERROR __main__.py:618: Check failed: linters/cmake-format/test_data/test2.in.cmake` as the lint output, so I'm not sure this is very actionable

TylerJang27: What are your thoughts on including `cmake-lint` here for the lint subcommand? We support turning on/off any number of linter subcommands, so we can start with some disabled by default.

linters/cmake-format/plugin.yaml
```diff
@@ -0,0 +1,30 @@
...
+        - name: fix
+          formatter: true
+          output: rewrite
+          success_codes: [0, 1]
+          run: cmake-format -i ${target}
+          in_place: true
+      tools: [cmake-format]
+      suggest_if: files_present
```
TylerJang27, line 26: Let's go ahead and change this to `config_present`. `files_present` is probably where we want to be long-term, but we usually dogfood new linters/let them marinate a bit before turning them on for a file basis.

linters/cmake-format/plugin.yaml
```diff
@@ -0,0 +1,30 @@
...
+      known_good_version: 0.6.13
```
TylerJang27, line 27: Can you also add the following:
```yaml
direct_configs:
  - cmake-format.yaml
  - cmake-format.json
  - cmake-format.py
  - .cmake-format.yaml
  - .cmake-format.json
  - .cmake-format.py
```
https://cmake-format.readthedocs.io/en/latest/configuration.html

linters/cmake-format/cmake-format.test.ts
```diff
@@ -0,0 +1,3 @@
+import { linterFmtTest } from "tests";
+
+linterFmtTest({ linterName: "cmake-format" });
```
TylerJang27, line 3: Depending on the outcome of the `lint` subcommand, it may be preferred to run `linterFmtTest` and `linterCheckTest` here.

TylerJang27 (commented):
Thanks for the PR! I left a few comments about some conventions, as well as some opportunity for discussion about the lint subcommand.

Could you also add this to the top-level README.md? We have a test that validates that all linters are included there. (`npm test tests/`)

TylerJang27:
@tmpsantos no rush, but if you'd like I can take over getting these changes landed. I would like to know your thoughts on the `cmake-format` lint subcommand though as a native user.

linters/cmake-format/plugin.yaml
tmpsantos: I see now, this was just me not fully understanding how this configuration worked. Only formatting should do.

Regarding `cmake-lint` I don't think it adds too much value as a tool. It mostly complains about lines being too long and bad indentation, which are corrected by the formatter.

README.md
```diff
@@ -45,6 +45,7 @@ trunk check enable {linter}
 | C#              | [dotnet-format]                                                                                                      |
 | CircleCI Config | [circleci]                                                                                                           |
 | Cloudformation  | [cfnlint], [checkov]                                                                                                 |
+| CMake           | [cmake-format]                                                                                                       |
```
TylerJang27, line 48: Can you also add a markdown link to https://cmake-format.readthedocs.io/en/latest/ on line 100 :)

TylerJang27 (approved):
LGTM, pending fixing the README.md

TylerJang27:
Thanks again for the PR! We'll get this out in the next release soon!

README.md
tmpsantos: Done!

PR #650, opened 2024-02-05
New linters/formatter for dart

TylerJang27:
Adds definitions for `format`, `fix`, and `analyze`. From a brief audit, users may prefer to remove individual subcommands but the full set is the best default.


debkanchan:
as a flutter dev

LET'S GOOOOOOOOOOOO

TylerJang27:
I will work on getting this landed this week. I need to do some auditing of the value of `dart analyze`, but otherwise the remaining work is relatively straightforward

laurit17 (approved):

PR #653, opened 2024-02-07
Add jq config query action

TylerJang27:
A simple `jq` wrapper in order to query information about the Trunk Config. Only runs when config is valid (as expected). Running `trunk run lookup .cli.version` returns
```
"1.19.1-beta.11"
```

Similarly, the following returns a linter's definition
```
trunk run lookup '.lint.definitions[] | select(.name == "eslint")'
```


TylerJang27:
Nevermind, `yq` is strictly better for this. I'll make a tool definition for `yq` when I get the chance

PR #655, opened 2024-02-07
Add tool: yq

TylerJang27:
Replaces #653. `yq` is the yaml equivalent of `jq`. Works really nicely to print a lint definition using:
`trunk config print | .trunk/tools/yq '.lint.definitions[] | select(.name == "eslint")'`, which could be streamlined via a Trunk Action if desired once tools are supported in Actions.


pat-trunk-io (approved):
lgtm

PR #660, opened 2024-02-13
Add `commitizen` action/tool

natescherer:
Hello!
This PR adds https://github.com/commitizen/cz-cli as both an action and a tool. I've named it `commitizen-cz-cli` instead of just `commitizen` to avoid confusion with the similarly-named, but entirely different Python-based https://github.com/commitizen-tools/commitizen.

This implements an action with `prepare-commit-msg` hook to allow commitizen to be injected into the normal `git commit` workflow, as well as as a tool to expose the `cz` CLI tool.

Did my best to manually test all of this as well as I could, wasn't really sure if you had standards for automated tests for actions/tools but I can implement if you can point me at a doc.

Please let me know if you have questions or would like fixes made!


actions/commitizen-cz-cli/README.md
```diff
@@ -0,0 +1,28 @@
+# Commitizen (cz-cli)
+
+https://github.com/commitizen/cz-cli
+
+## Simple configuration
+
+Under the default configuration, Commitizen will pull its configuration from commitlint config files, as there is a high likelihood both tools will be used together. The below sample configuration will apply the [@commitlint/config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional) standard for both Commitizen and commitlint.
+
+1. `echo "{"path": "@commitlint/cz-commitlint"}" > .czrc`
```
det, line 9:
```suggestion
1. `echo '{"path": "@commitlint/cz-commitlint"}' > .czrc`
```
When I tried this my shell ate the inner quotes. This should fix it. Or maybe just use code blocks for all of these that people can copy-paste? ie:
Place the following in `.czrc`:
```
"path": "@commitlint/cz-commitlint"}
```

actions/commitizen-cz-cli/plugin.yaml
```diff
@@ -0,0 +1,24 @@
+version: 0.1
+actions:
+  definitions:
+    - id: commitizen-cz-cli
+      display_name: Commitizen (cz-cli)
+      description: Assistant for writing commit messages that obey a standard
+      runtime: node
+      packages_file: package.json
```
det, line 8: I've opened an internal feature request to support tool dependencies in actions. It would be nice not to need the `package_file` here.

det (commented):
Thank you for your contribution! I tested everything locally and it works for me excepting some issues with my shell eating quotes that I left a comment on.

Also, it looks like this `commitizen` is much more popular than the Python project. I think it's better if it uses that name, but I will defer to you on that.

natescherer:
Ahh yeah that quoting was a miss on my part. Will resolve later tonight. However, I've subsequently realized my naming has an overall problem: the Python package *also* calls its CLI cz, so I'm not actually avoiding any confusion here. Thoughts on renaming to commitizen-node, perhaps? Or if you think just conmitizen is fine, I'm good with that too.

det:
Thanks. I think `commitizen` is best for this. They own that top-level name on GitHub and have 8x the number of stars of the other project.

actions/commitizen-cz-cli/plugin.yaml
natescherer: Sounds good!

natescherer:
Okay, changes made and tested locally. Let me know if there is anything else!

det (approved):
LGTM, Thank you!

natescherer:
Saw check was failing due to the README not passing Prettier, fixed now.

PR #661, opened 2024-02-13
Better markdownlint config

dapirian:
They support a prettier-friendly-config themselves now


det (approved):

TylerJang27:
@dapirian this looks like either the snapshots will need to be updated or the config will need some tuning to turn off `line_length` and `blank_lines`

PR #668, opened 2024-02-20
Add `pwsh` Tool  and `psscriptanalyzer` Linter

natescherer:
Hello!
This PR adds support for PowerShell via creating a `pwsh` tool and a linter/formatter `psscriptanalyzer` that uses it.

A few notes about things that could be improved in this, but I ran into limitations on:

* Tools seem to require shims, so I had to make a useless shim to PSScriptAnalyzer.psd1, which isn't executable but I didn't know what else to do
* I have to ad-hoc install the ConvertToSARIF module as part of the lint.ps1 script, as it's necessary to convert the output and I couldn't figure any other way to install it.
  * If there is a way to pass in the root directory of a tool that a linter depends on, ConvertToSARIF could be added as a tool and the module imported similarly to how the PSScriptAnalyzer module itself is imported

This is my first time adding a linter, so let me know if I did anything wrong or if there are additional things I need to do.

Thanks!

### DIFF
--- a/linters/psscriptanalyzer/lint.ps1
+++ b/linters/psscriptanalyzer/lint.ps1
@@ -28,6 +28,6 @@ if (Test-Path "PSScriptAnalyzerSettings.psd1") {
     }
 }
 
-Invoke-ScriptAnalyzer @Analyzer Splat | ConvertTo-SARIF -FilePath $OutputPath
+Invoke-ScriptAnalyzer @AnalyzerSplat | ConvertTo-SARIF -FilePath $OutputPath
 
 Remove-Item -Path (Join-Path "TEMP:" "ConvertToSARIF") -Force
--- a/linters/psscriptanalyzer/lint.ps1
+++ b/linters/psscriptanalyzer/lint.ps1
@@ -1,21 +1,33 @@
 param(
-  [Parameter(Mandatory = $true)]
-  [string]$ModuleDir,
-  [Parameter(Mandatory = $true)]
-  [string]$FilePath,
-  [Parameter(Mandatory = $true)]
-  [string]$OutputPath
+    [Parameter(Mandatory = $true)]
+    [string]$ModuleDir,
+    [Parameter(Mandatory = $true)]
+    [string]$FilePath,
+    [Parameter(Mandatory = $true)]
+    [string]$OutputPath
 )
 
 # Import PSScriptAnalyzer module
 Import-Module -Name (Join-Path $ModuleDir PSScriptAnalyzer.psd1)
 
 # Download and import ConvertToSARIF module
+# NOTE: This could be converted to a tool once there is a way to pass in root directories of tools other than the linter itself
 $Version = "1.0.0"
 $ImportPath = Join-Path "TEMP:" "ConvertToSARIF" $Version
 Save-Module -Name ConvertToSARIF -Path "TEMP:" -RequiredVersion $Version
 Import-Module -Name (Join-Path $ImportPath "ConvertToSARIF.psd1")
 
-Invoke-ScriptAnalyzer -Path $FilePath -Settings PSScriptAnalyzerSettings.psd1 -Fix | ConvertTo-SARIF -FilePath $OutputPath
+$AnalyzerSplat = @{
+    Path = $FilePath
+    Fix  = $true
+}
+
+if (Test-Path "PSScriptAnalyzerSettings.psd1") {
+    $AnalyzerSplat += @{
+        Settings = "PSScriptAnalyzerSettings.psd1"
+    }
+}
+
+Invoke-ScriptAnalyzer @Analyzer Splat | ConvertTo-SARIF -FilePath $OutputPath
 
 Remove-Item -Path (Join-Path "TEMP:" "ConvertToSARIF") -Force
--- a/linters/psscriptanalyzer/lint.ps1
+++ b/linters/psscriptanalyzer/lint.ps1
@@ -1,19 +1,21 @@
 param(
-    [Parameter(Mandatory = $true)]
-    [string]$FilePath,
-    [Parameter(Mandatory = $true)]
-    [string]$OutputPath
+  [Parameter(Mandatory = $true)]
+  [string]$ModuleDir,
+  [Parameter(Mandatory = $true)]
+  [string]$FilePath,
+  [Parameter(Mandatory = $true)]
+  [string]$OutputPath
 )
 
-# Install PSScriptAnalyzer module, if needed
-if ($null -eq (Get-InstalledModule -Name PSScriptAnalyzer -ErrorAction SilentlyContinue -WarningAction SilentlyContinue)) {
-    Install-Module -Name PSScriptAnalyzer -Force
-}
+# Import PSScriptAnalyzer module
+Import-Module -Name (Join-Path $ModuleDir PSScriptAnalyzer.psd1)
 
-# Install ConvertToSARIF module, if needed
-if ($null -eq (Get-InstalledModule -Name ConvertToSARIF -ErrorAction SilentlyContinue -WarningAction SilentlyContinue)) {
-    Install-Module -Name ConvertToSARIF -Force
-}
-Import-Module -Name ConvertToSARIF -Force
+# Download and import ConvertToSARIF module
+$Version = "1.0.0"
+$ImportPath = Join-Path "TEMP:" "ConvertToSARIF" $Version
+Save-Module -Name ConvertToSARIF -Path "TEMP:" -RequiredVersion $Version
+Import-Module -Name (Join-Path $ImportPath "ConvertToSARIF.psd1")
 
 Invoke-ScriptAnalyzer -Path $FilePath -Settings PSScriptAnalyzerSettings.psd1 -Fix | ConvertTo-SARIF -FilePath $OutputPath
+
+Remove-Item -Path (Join-Path "TEMP:" "ConvertToSARIF") -Force
--- a/linters/psscriptanalyzer/lint.ps1
+++ b/linters/psscriptanalyzer/lint.ps1
@@ -1,0 +1,19 @@
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$FilePath,
+    [Parameter(Mandatory = $true)]
+    [string]$OutputPath
+)
+
+# Install PSScriptAnalyzer module, if needed
+if ($null -eq (Get-InstalledModule -Name PSScriptAnalyzer -ErrorAction SilentlyContinue -WarningAction SilentlyContinue)) {
+    Install-Module -Name PSScriptAnalyzer -Force
+}
+
+# Install ConvertToSARIF module, if needed
+if ($null -eq (Get-InstalledModule -Name ConvertToSARIF -ErrorAction SilentlyContinue -WarningAction SilentlyContinue)) {
+    Install-Module -Name ConvertToSARIF -Force
+}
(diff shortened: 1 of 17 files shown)


linters/psscriptanalyzer/lint.ps1
```diff
@@ -0,0 +1,33 @@
...
+
+# Import PSScriptAnalyzer module
+Import-Module -Name (Join-Path $ModuleDir PSScriptAnalyzer.psd1)
+
+# Download and import ConvertToSARIF module
+# NOTE: This could be converted to a tool once there is a way to pass in root directories of tools other than the linter itself
+$Version = "1.0.0"
+$ImportPath = Join-Path "TEMP:" "ConvertToSARIF" $Version
+Save-Module -Name ConvertToSARIF -Path "TEMP:" -RequiredVersion $Version
+Import-Module -Name (Join-Path $ImportPath "ConvertToSARIF.psd1")
```
TylerJang27, line 18: This is a good callout about tool dir resolution, and it's something we will look into adding in the future.

With that being said, the code as you have written here doesn't work for me for a couple reasons:
1. This doesn't work in parallel since the installations (and removal) will clobber each other
2. Not imperative, but it would be better to not have to reinstall ConvertToSarif on every execution
3. `TEMP` doesn't resolve correctly on my machine. I think we would want to use something like `$env:TEMP` and default it to `/tmp`.

We have a couple options to address this:
1. Does powershell have a convenient way to search for `ConvertToSarif.psd1` in the PATH? Tools work under the hood by just exporting the installation PATH for the linter.
2. Don't install `ConvertToSarif`, and instead use a Trunk custom parser ([example](https://github.com/trunk-io/plugins/blob/main/linters/sqlfluff/sqlfluff_to_sarif.py)). The output format looks CSV-like so that might be an option to eliminate the dependency
4. We expose the environment variable `${shared_cachedir}` for cases like this, but then the script would have to be thread-safe.

What are your thoughts on these options?

linters/psscriptanalyzer/plugin.yaml
```diff
@@ -0,0 +1,43 @@
...
+      known_good_version: 1.21.0
+lint:
+  definitions:
+    - name: psscriptanalyzer
+      files: [powershell]
+      tools: [psscriptanalyzer, pwsh]
+      commands:
+        - name: Invoke-ScriptAnalyzer
+          run: |
+            pwsh -f ${cwd}/lint.ps1 -ModuleDir ${linter} -FilePath ${target} -OutputPath ${tmpfile}
```
TylerJang27, line 24: Can you inline this? Otherwise yaml actually treats there as being an implicit new lint at the end. I think that is the source of the test failure in CI.

linters/psscriptanalyzer/plugin.yaml
```diff
@@ -0,0 +1,43 @@
...
+          output: sarif
+          read_output_from: tmp_file
+          batch: false
+          success_codes: [0]
+        - name: Invoke-Formatter
+          run: |
+            pwsh -f ${cwd}/format.ps1 -ModuleDir ${linter} -FilePath ${target}
```
TylerJang27, line 31: Also inline

TylerJang27 (commented):
Thanks for the PR! This is a great addition to our toolkit, and I appreciate the thoroughness of your diff. I've left a couple comments, particularly in relation to the sarif conversion. Let me know what you think and I can help you drive this over the finish line.

linters/psscriptanalyzer/lint.ps1
natescherer: Did not think about this needing to run in parallel, good callout. Also `TEMP:` (important to include the colon) is a PowerShell virtual drive that always points to an appropriate temp folder for the platform it is on (and is only available inside pwsh itself).

Comments on the options:
1. So I thought about trying to import `ConvertToSARIF.psd1`, but tool shims aren't actually symlinks, at least on my install, they are a bash wrapper script that then goes and points at the actual tool. Is there a way to have it be a symlink? If so, option 1 might work.
2. Certainly doable, but as it's the most work I'd prefer to try another option if I can.
3. Interesting, I wasn't aware of that variable. Can you elaborate more on what you mean by needing to make sure the script is thread safe? PowerShell is single-threaded by default.

linters/psscriptanalyzer/plugin.yaml
natescherer: Done.

linters/psscriptanalyzer/plugin.yaml
natescherer: I removed the pipe but then (I assume) Prettier went and pushed the statement onto the next line anyway. Hopefully that is okay?

TylerJang27: I think that should be fine. The tests matched my observations locally so if those pass we should be good

linters/psscriptanalyzer/lint.ps1
TylerJang27: Thanks for the `TEMP:` explanation. I was running into some issues but it might have been a different problem. I will give it another try when we get the sarif bit figured out.

1. It isn't the shims that we would actually be leveraging. When you specify `tools: [psscriptanalyzer, pwsh]`, it implicitly adds those tools/downloads to the `PATH` when executing the linter. You can see this if you run `trunk check --verbose --force linters/psscriptanalyzer/test_data/check.in.ps1` after enabling and then inspect the yaml file. I'm not very familiar with how powershell resolves commands from the PATH, but if we can search from there we could make this a tool and Trunk will handle the installation.
2. Definitely agree.
3. It's not something we document since it's mainly an implementation detail, but it might work here. I'm not aware of the powershell options for something like this, but a naive implementation would be:

---

1. Check if installed `${shared_cachedir}/ConvertToSARIF` exists. If it does, good to execute.
2. If not, make a file `${shared_cachedir}/ConvertToSARIF.lock` and lock it.
3. Download/install the module to `${shared_cachedir}/ConvertToSARIF`
4. Unlock `${shared_cachedir}/ConvertToSARIF.lock`
5. Execute the linting

Or something like [this](https://stackoverflow.com/questions/63727114/file-lock-mechanism-using-powershell)

natescherer: Thanks for the fast and detailed explanations. I'll pursue options 1, 3, and 2 in that order, as I think all are viable.

TylerJang27: Thanks again for implementing this. I'm happy to help more with any minutiae, but my powershell knowledge is limited. Looking forward to getting this working!

natescherer: Option 1 turned out to be quite easy now that I know the downloads are in the PATH. I like this a lot better.

natescherer:
Think we're close. I'll see if I can figure out why tests on Windows are behaving differently.

linters/psscriptanalyzer/plugin.yaml
natescherer: Other changes I made made this line short enough now, anyway.

linters/psscriptanalyzer/plugin.yaml
```diff
@@ -7,27 +7,37 @@ downloads:
...
+    - name: converttosarif
+      download: converttosarif
+      shims: [ConvertToSARIF.psd1]
+      known_good_version: 1.0.0
 lint:
   definitions:
     - name: psscriptanalyzer
       files: [powershell]
-      tools: [psscriptanalyzer, pwsh]
+      tools: [psscriptanalyzer, converttosarif, pwsh]
```
TylerJang27, line 31: As a minor convention change, please change this to:
```yaml
main_tool: psscriptanalyzer
tools: [converttosarif, pwsh]
```

This is just more explicit so that the version that is enabled applies to `psscriptanalyzer`

linters/psscriptanalyzer/plugin.yaml
```diff
@@ -0,0 +1,52 @@
...
+        - name: Invoke-Formatter
+          run: pwsh -f ${cwd}/format.ps1 -FilePath ${target}
+          output: rewrite
+          formatter: true
+          in_place: true
+          batch: false
+          success_codes: [0]
+      direct_configs: [PSScriptAnalyzerSettings.psd1]
+      suggest_if: config_present
+      run_linter_from: workspace
```
TylerJang27, line 48: nit: can you remove this line, since all linters are implicitly run from the workspace, and we prefer the new `run_from:` syntax anyway

TylerJang27 (commented):
Thanks for the change. Almost there! I'll take a closer look and try to repro the Windows issue tomorrow. I suspect it may be a problem with the testing framework...

natescherer:
> Thanks for the change. Almost there! I'll take a closer look and try to repro the Windows issue tomorrow. I suspect it may be a problem with the testing framework...

Yeah, I've been completely unable to figure out how to get the tests to run on my local Windows box so I'm not gonna be much help on that one. Could it be that maybe the windows git is doing autocrlf?

linters/psscriptanalyzer/plugin.yaml
natescherer: Done. FYI, I think this doc needs updated to reflect that syntax: https://docs.trunk.io/check/configuration/custom-linters#additional-tool-dependencies

linters/psscriptanalyzer/plugin.yaml
natescherer: Done.

TylerJang27:
@natescherer I figured out the issue. For whatever reason, PSScriptAnalyzer auto-formats with carriage returns on Windows. The easiest path forward to deal with this is just to make the following changes to the test file to be platform-agnostic. You can also delete your `*_custom_settings.ps1` files and delete/regenerate the snapshots.

```
import path from "path";
import { customLinterCheckTest, linterFmtTest } from "tests";
import { TrunkLintDriver } from "tests/driver";
import { TEST_DATA } from "tests/utils";

// Run with `-y` since on Windows Invoke-Formatter will add carriage returns, and for testing's sake it's easier to just apply.
const checkArgs = `${path.join(TEST_DATA, "check.in.ps1")} -y`;

// Run tests with default rules
linterFmtTest({ linterName: "psscriptanalyzer", namedTestPrefixes: ["format"] });
customLinterCheckTest({ linterName: "psscriptanalyzer", testName: "check", args: checkArgs });

// Create a PSScriptAnalyzerSettings.psd1 for further testing
const preCheck = (driver: TrunkLintDriver) => {
  driver.writeFile(
    ".trunk/configs/PSScriptAnalyzerSettings.psd1",
    `@{
ExcludeRules = @("PSAvoidUsingWriteHost","AvoidTrailingWhitespace")
Rules = @{PSAvoidSemicolonsAsLineTerminators = @{Enable = $true}}
}`,
  );
};

// Run tests with custom settings
linterFmtTest({
  linterName: "psscriptanalyzer",
  namedTestPrefixes: ["format"],
  preCheck,
});
customLinterCheckTest({
  linterName: "psscriptanalyzer",
  testName: "check_custom_settings",
  args: checkArgs,
  preCheck,
});
```

TylerJang27:
Hi @natescherer, we'd love to get this merged soon! Do you want any remaining help getting this landed or would you like me to take over this PR?

natescherer:
Sorry for the delay. I thought it smelled like a line ending issue. Just made your suggested changes, let's see if this passes CI...

TylerJang27 (approved):
LGTM. Thanks for the PR! We'll get this into the next release

PR #676, opened 2024-02-27
Update toolbox definiton in plugins

EliSchleifer:
toolbox definition was pinned to a specific version / should have been a range

### DIFF
--- a/linters/trunk-toolbox/plugin.yaml
+++ b/linters/trunk-toolbox/plugin.yaml
@@ -2,28 +2,28 @@ version: 0.1
 
 downloads:
   - name: trunk-toolbox
-    version: 0.2.0
+    version: 0.3.0
     downloads:
-      - version: 0.2.0
-        os:
+      - os:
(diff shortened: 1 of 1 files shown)


linters/trunk-toolbox/plugin.yaml
```diff
@@ -2,31 +2,31 @@ version: 0.1
...
       shims: [trunk-toolbox]
-      known_good_version: 0.2.0
+      known_good_version: 0.3.0
 lint:
   definitions:
     - name: trunk-toolbox
       tools: [trunk-toolbox]
       files: [ALL]
-      known_good_version: 0.2.0
+      known_good_version: 0.3.0
```
TylerJang27, line 26: IIRC this line is superfluous.

Also nit: we can change `tools: [trunk-toolbox]` to `main_tool: trunk-toolbox` just to be more explicit

TylerJang27 (approved):
Minor cleanup

PR #687, opened 2024-03-06
fix(osv-scanner): include package name in message

puzzler7:
This PR:
- adds the package name to the message if it's not already included in the output (`Current version of "pkg" is vulnerable` instead of `Current version is vulnerable`).
- fixes line numbers in lock file formats that have the name on one line followed by the version on the next (such as Cargo.lock, package-lock.json, etc)
- Updates the osv-scanner plugin definition to download osv-scanner >1.7.0 as they changed their download filename format


TylerJang27 (approved):
LGTM

PR #689, opened 2024-03-07
Update prettier parsing to respect syntax errors even if they're not the first line of output

dapirian:


pat-trunk-io (approved):
nit add test

PR #709, opened 2024-03-21
Rename poorly named variable

det:


TylerJang27 (approved):
🙂

linters/tofu/terraform.test.ts
TylerJang27, line 1: Looks like this file was included by mistake and should be removed

det: already deleted

TylerJang27 (approved):

PR #718, opened 2024-03-27
Add Poetry actions

TylerJang27:
Dupe of #229. Adds poetry Trunk Actions to match parity with [pre-commit](https://python-poetry.org/docs/pre-commit-hooks/). Note that the non-hermetic story is much cleaner at time of writing, with the hermetic story blocked by not being able to unset `VIRTUAL_ENV` in the Action and Tool. In other words, for now recommended behavior requires manual [Poetry installation](https://python-poetry.org/docs/#installation).

We've added a tool test and several actions tests


dapirian (approved):
lgtm!

PR #726, opened 2024-04-03
Update commitlint v17 -> v19

rriski:
v19 contains breaking changes, see:
- https://github.com/conventional-changelog/commitlint/blob/master/CHANGELOG.md#1900-2024-02-27

### DIFF
--- a/actions/commitlint/commitlint.test.ts
+++ b/actions/commitlint/commitlint.test.ts
@@ -4,8 +4,8 @@ import { TrunkActionDriver } from "tests/driver";
 
 const preCheck = (driver: TrunkActionDriver) => {
   driver.writeFile(
-    "commitlint.config.js",
-    "module.exports = {extends: ['@commitlint/config-conventional']}",
+    "commitlint.config.mjs",
+    "export default {extends: ['@commitlint/config-conventional']}",
   );
 };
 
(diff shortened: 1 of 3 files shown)


actions/commitlint/commitlint.test.ts
```diff
@@ -4,8 +4,8 @@ import { TrunkActionDriver } from "tests/driver";
 
 const preCheck = (driver: TrunkActionDriver) => {
   driver.writeFile(
-    "commitlint.config.js",
-    "module.exports = {extends: ['@commitlint/config-conventional']}",
+    "commitlint.config.mjs",
+    "export default {extends: ['@commitlint/config-conventional']}",
```
rriski, line 8: Not sure how renaming the file and using `export default` should be handled by trunk when upgrading to this version. I.e. how to handle breaking changers in tools supported by trunk?

rriski: I suppose its up to the user to configure the commitlint config file. I’ll need to update the commitlint readme here: https://github.com/trunk-io/plugins/tree/main/actions/commitlint

TylerJang27: Thanks for the PR @rriski and for bringing this to our attention! I'll speak with the team about if there is any better way we can handle the update process for users. I think we will probably want to go ahead and land this, and also update the readme instructions, as you mentioned, but I'll follow-up again shortly.

TylerJang27 (approved):
Thanks for the PR! I will call this out in our next release as a breaking change

PR #730, opened 2024-04-05
Cleanup install on ubuntu runners

TylerJang27:
Fix [failures](https://github.com/trunk-io/plugins/actions/runs/8566927464/job/23477663625) introduced by php setup. Successful [run](https://github.com/trunk-io/plugins/actions/runs/8575992769/job/23506162438?pr=730)


.github/actions/linter_tests/action.yaml
```diff
@@ -44,6 +44,11 @@ runs:
 
         case "$RUNNER_OS" in
           Linux)
+            # Cleanup any existing apt locks
+            sudo killall apt-get || true
+            sudo rm /var/lib/apt/lists/lock || true
```
TylerJang27, line 49: There was a 2-day old apt-get process that still had the lock on one of the runners. This should rarely/never happen, so just clean up as part of the action runner process, because it will fail otherwise anyway

PR #733, opened 2024-04-08
Don't cache pyright results

det:
It needs access to all files in the repository.


dapirian (approved):

TylerJang27:
This is already the case

PR #734, opened 2024-04-09
Upgrade ruby-build

det:
Tested locally after deleting ruby and ruby-build from my cache.


TylerJang27:
Converting to draft until we get to the bottom of the original user's issue

PR #735, opened 2024-04-09
Support and migrate to eslint@9.0.0

TylerJang27:
[Eslint@9.0.0](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/) was released and with it a major change to how eslint is configured. The [migration process](https://eslint.org/docs/latest/use/configure/migration-guide) is nontrivial and may not be an option/priority for every project. Additionally, upgrading will loudly fail unless you have accounted for the config. Accordingly, I have done the following:

- Make `direct_configs` a property of `commands`, allowing us to version the configs that are applicable for linter
    - Users will only be upgraded/init'd/enabled at compatible versions with their config
- Modify the `eslint` tests to handle pre- and post- eslint9
- Upgrade this repo to use an `eslint9` _flat config_
- Perform an audit and address a handful of the new diagnostics coming from `eslint9`

This will requires a minimum version bump to support. _Most of the files changes are lockfiles._


TylerJang27:
Ultimately, there are 2 existing issues warranting this change (or a similar one):
1. LUV cannot safely upgrade a user on eslint@8.x to eslint@9.x, since it will 100% break said user, even with the environment variable override that doesn't really work
2. If we detect an eslint config, we need to know which version of eslint to enable at. Having an `@package` would alleviate this issue

TylerJang27:
@det Plan for release since older CLI versions will naively respect LUV, AND also won't have the direct_configs update here:
1. Land this
2. Follow-up with a patch that omits eslint from LUV uploading
3. Release plugins
4. Let it marinate for ~1 week so that the newest LUV will stay at 8.57.0
5. Revert the patch so that LUV will have the correct newest validated eslint version

The patch is only necessary for extra caution, in case someone is on a new CLI version but not a new plugins version (or on an old CLI version).

det (approved):

PR #740, opened 2024-04-11
Add prettier support for postcss

### LINKED ISSUE #738: Request: add 'postcss' file support to prettier plugin
e.g. sveltekit uses `*.postcss` files to trigger automatic postcss processing when using the default vite preprocessor. Prettier supports formatting these files.

TylerJang27:
Tested manually on a toy example that prettier formats these correctly. Using [postcss extension](https://marketplace.visualstudio.com/items?itemName=csstools.postcss) as reference. Fixes #738

### DIFF
--- a/linters/plugin.yaml
+++ b/linters/plugin.yaml
@@ -412,6 +412,12 @@ lint:
       extensions:
         - png
 
+    - name: postcss
+      extensions:
+        # TODO(Tyler): Do we want to include css here as well?
+        - pcss
+        - postcss
+
     - name: powershell
       extensions:
         - ps1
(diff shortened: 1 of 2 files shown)


linters/plugin.yaml
```diff
@@ -412,6 +412,12 @@ lint:
       extensions:
         - png
 
+    - name: postcss
+      extensions:
+        # TODO(Tyler): Do we want to include css here as well?
```
dapirian, line 417: don't think so? we have a `css` filetype already?

dapirian (approved):

linters/plugin.yaml
TylerJang27: Yeah, although there's an intersection. It may be more ergonomic to include it here with an inherits, but probably best to keep separate for extensibility

PR #741, opened 2024-04-11
Merge bazel filetypes and update to include MODULE.bazel

dapirian:


TylerJang27 (approved):
LGTM. Do you know if there was a historical reason why we separated them out?

dapirian:
> LGTM. Do you know if there was a historical reason why we separated them out?

Pedanticness. I ended up leaving the separation for backwards compat but used the `inherits` feature to consolidate them forward-looking.

det (approved):

yada.pyi
TylerJang27, line 1: Remove plz

dapirian: oops thanks

PR #744, opened 2024-04-15
Not suggest terrascan

yu-iskw:
terrascan is no longer actively maintained. So, it would be good not to suggest even if related files exist.

### DIFF
--- a/linters/terrascan/plugin.yaml
+++ b/linters/terrascan/plugin.yaml
@@ -22,7 +22,7 @@ lint:
     - name: terrascan
       tools: [terrascan]
       known_good_version: 1.18.1
-      suggest_if: files_present
+      suggest_if: never
       commands:
         - name: lint
           output: sarif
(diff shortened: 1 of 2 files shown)


TylerJang27:
Hi @yu-iskw! Thanks for flagging this: I'll discuss with the team if we want to go ahead and turn this off by default for everyone. A couple questions:
- On what basis were you saying it's no longer maintained? It looks like their last release was in [March](https://github.com/tenable/terrascan/releases/tag/v1.19.1) (although I do see the substantial backlog of issues)
- Were you encountering a lot of linter failures when running `terrascan` in your repo? If so, could you include the failure outputs? It's possible we can modify our integration to improve the failure rate.

yu-iskw:
According to [the commit stats](https://github.com/tenable/terrascan/graphs/commit-activity) of terrascan, I don't think it is actively developed. 

image

For instance, terrascan doesn't support newer versions of terraform. For instance, it doesn't support the `import` block. I encountered violations caused by the less active maintenance. So, I decided not to use terrascan. 

https://github.com/tenable/terrascan/issues/1590

TylerJang27:
I talked this one over with the team, and we agreed to disable it for now. Thanks for the PR!

TylerJang27 (approved):

PR #757, opened 2024-04-26
Adding some of the most common php tooling 

matthewbaggett:
* phpunit for running unittests
* paratest for running the same unitests in parallel
* php-cs-fixer for enforcing formatting. 

php-cs-fixer could be considered to be promoted to being a linter as soon as I work out how to parse its output and feed it through a post-processor like phpstan is.


laurit17:
Hi Matthew, thank you so much for taking the lead here on integrating these tools. In order to get them to pass, would you mind adding health checks and unit tests? You can look at These are very simple. Let me/us know if you have any trouble.

I'd do it myself but you're using a fork, not a branch :)

example: https://github.com/trunk-io/plugins/blob/main/tools/clangd/plugin.yaml#L20

EliSchleifer:
@matthewbaggett - thanks for opening this pull request. would love to get these landed. do you need more guidance to get the tests in place?

TylerJang27:
@matthewbaggett I've gone and added the dependencies and tests to get this working. Can you try `php-cs-fixer` out and make sure it conforms to your original needs/idiomatic usage? Otherwise I will land this tomorrow.

TylerJang27 (approved):

PR #762, opened 2024-04-30
Cleanup contributing copy and streamline repo styling

TylerJang27:
1. Updates and streamlines the copy for `README.md`, `CONTRIBUTING.md`, and `tests/README.md` to only focus on the important points and self-reference each other more clearly
    - Also underscores the autogen feature (requires a daemon launch)
2. Replaces some of the repo tests with a `node` linter `definition-checker` that enforces stylistic choices in the repo and recommends conventions
    - Validates no linters/tools/actions are explicitly enabled
    - Ensures linters have `description` and `suggest_if` set
    - Ensures all linters have a relevant test file
3. Fixes those conventions with a few places that fell through the cracks
4. Extracts out the codegen into separate template files and reference them from `CONTRIBUTING.md`


CONTRIBUTING.md
```diff
@@ -32,43 +38,24 @@ plugins:
...
-
 ## Linters
 
 To add a new linter:
 
 1. Run `trunk check` to start up Trunk in the background.
-2. Create a directory inside `linters/` with the name of your new linter.
-3. Inside this new directory, create the following structure. Most of these files will be
-   automatically created for you:
+2. Run `mkdir linters/<my-linter>` to start. This should autopopulate with a sample
```
gewenyu99, line 55: Do you have to run this before check finishes? I think this is the part I missed 😆

gewenyu99:
Nice :D ❤️ This would definitely be helpful for the next contributor.

I think the only other thing that took me some time to figure out was all the temporary folders generated when a tool's downloaded/when tests are running locally.

They were useful to check when things didn't act like I'd expect. Would be nice if smwhere in contributing we can point in the general direction. I happened to find them just by some weird `grep`s I tried running.

CONTRIBUTING.md
TylerJang27: Really `trunk check` here is just a shorthand for launching our daemon, which listens for file changes in order to run [Trunk Actions](https://github.com/trunk-io/plugins/blob/main/.trunk/trunk.yaml#L47-L54). `trunk check` should finish almost immediately if you're starting from a fresh checkout.

Running `mkdir` after that (1-2 seconds after the `trunk check` command starts) should be sufficient. But if you ran something like `code linters/vale/plugin.yaml`, the [script](https://github.com/trunk-io/plugins/blob/main/repo-tools/linter-test-helper/generate#L80-L85) won't autogen since it only checks empty directories :)

TylerJang27:
@gewenyu99 Thanks for the feedback! I'll update this to make sure that info is more discoverable!

.trunk/trunk.yaml
```diff
@@ -20,7 +20,27 @@ plugins:
...
+      description: Checks plugin.yaml files for repo best practices
+      files: [plugin.yaml]
+      runtime: node
+      extra_packages:
+        - ts-node
+        - yaml
+      commands:
+        - name: check
+          run: ts-node ${workspace}/repo-tools/definition-checker/check.ts ${target}
+          batch: true
```
TylerJang27, line 37:
```suggestion
          batch: true
          sandbox_type: expanded
```

CONTRIBUTING.md
det: We have `trunk daemon start` that just stats the daemon.

det (approved):

PR #774, opened 2024-05-08
Setup Flaky Test monitoring

TylerJang27:
Adds an upload step so that we upload all our test results. Validated on [webapp](https://app.trunk-staging.io/trunk-staging-org/trunk-io/plugins/test-analytics?statuses=Broken%2CFlaky%2CNew). The next step once the results seem to be reliable would be to turn on test quarantining and run with that action (note that it does not support Windows).

### DIFF
--- a/.github/actions/action_tests/action.yaml
+++ b/.github/actions/action_tests/action.yaml
@@ -73,8 +73,8 @@ runs:
         PLUGINS_TEST_CLI_PATH: ${{ env.CLI_PATH }}
 
     - name: Upload results
-      # TODO(Tyler): Add upload on Windows once the action supports it.
-      if: "!cancelled() && runner.os != 'Windows'"
+      # TODO(Tyler): Add upload on MacOS/Windows once the action supports it.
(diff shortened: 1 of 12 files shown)


.github/actions/action_tests/action.yaml
```diff
@@ -64,10 +64,22 @@ runs:
...
         org: trunk-staging-org
         run:
           npm test ${{ inputs.append-args }} ${{ env.PLATFORM_APPEND_ARGS }} --passWithNoTests --ci
       env:
         PLUGINS_TEST_CLI_VERSION: ${{ inputs.cli-version }}
         PLUGINS_TEST_CLI_PATH: ${{ env.CLI_PATH }}
+
+    - name: Upload results
+      # TODO(Tyler): Add upload on Windows once the action supports it.
+      if: "!cancelled() && runner.os != 'Windows'"
```
gnalh, line 77: We also don't support Darwin.

gnalh (approved):

PR #801, opened 2024-06-07
[feat] add trunk fmt pre-push hook

Ryang20718:
A user may have multiple commits they want to add in series. whilst trunk fmt is fast, it may not be instantaneous. would be nice to have an option to run prepush


TylerJang27:
Hi @Ryang20718, thanks for the PR! I think that this action will not work as intended, however. `GIT_INDEX_FILE` is something that comes from git as an input to a pre-commit hook. `trunk-fmt-pre-commit` uses that to modify the index in place, so that the correct formatting of your files gets committed (and you don't have to run `git commit` again).

For a `pre-push` hook, there's no such construct. If you still want to add something like this, you could do the following with a wrapper:
1. Check that the git workspace is clean
2. Run trunk fmt (probably just with `--trigger=pre-push`)
3. Add the changed files and commit them
4. Verify that the correct ref is still being pushed

Let me know what you think, it's possible that there's also another option to smooth out your workflow; what formatter is usually the slowest to complete?

Ryang20718:
@TylerJang27 ready for a re-review!

TylerJang27 (commented):
Can you add a comment explaining this action, since it's a slight departure from idiomatic usage?

TylerJang27:
@Ryang20718, I'd like to get this landed this week if possible. I can add the comment, but I just wanted to check in with you, have you been using this and has the experience worked well for you?

Ryang20718:
Apologies, missed the notification here!

We eventually ended up vendoring our own. We had users complain that this didn't work for certain solutions, i.e vscode PR's, emacs, etc so we had to add a lot of env variables to disable features for users

TylerJang27:
@Ryang20718 gotcha. If you have any specific asks for what that feature (in)compatibility was, please let us know!

Ryang20718:
if you create a pull request using vscode's built in git integration, there's no way to skip the fmt checks (for some users they didn't need to merge main so they wanted to skip)

TylerJang27:
We have `interactive: true` which should solve that. We have a distinction between `trunk-check-pre-push` and `trunk-check-pre-push-always`, where the latter will run using VSCode's built-in SCM and the former will not. This _should_ be identical with the fmt command, but I can try it out at some point to see how it behaves

PR #802, opened 2024-06-07
Make newer eslint command versionless

TylerJang27:
This is for clarity and so overrides continue to work without breaking config


TylerJang27:
Decided we don't want to do this, we want to change how overrides work

PR #844, opened 2024-08-12
(Chore): Fix eslint rules

TylerJang27:
Fixes a failure with eslint because of updates in some of the rules dependencies. Also does some upkeep of old ignores and rules.

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -44,11 +44,6 @@ module.exports = [
       "no-unused-vars": "off",
       "no-use-before-define": "off",
       "no-useless-constructor": "off",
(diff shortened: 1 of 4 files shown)


eslint.config.cjs
puzzler7, line 1: Sanity check, should this get put in the configs repo?

puzzler7 (approved):
LGTM pending sanity

eslint.config.cjs
TylerJang27: No, eslint pre- and post- 9 are very different, including with their supported configs files. This is our only repo that is fully on eslint post-9

PR #887, opened 2024-10-11
(Feat): Add tool grpcui

TylerJang27:
Add tool [grpcui](https://github.com/fullstorydev/grpcui). It's available as a download or go install, but a download here allows us to query for latest


katchao (approved):
Very cool thank you

PR #902, opened 2024-10-30
Update trunk toolbox config to include additional direct config value - this better supports toolbox logging

EliSchleifer:


linters/trunk-toolbox/plugin.yaml
```diff
@@ -33,7 +33,10 @@ lint:
           output: sarif
           batch: true
           success_codes: [0]
+          disable_upstream: false
           read_output_from: tmp_file
+          direct_configs: [toolbox.toml, log4rs.yaml]
```
TylerJang27, line 38: nit: should be `affects_cache`

TylerJang27 (approved):
LGTM

linters/trunk-toolbox/plugin.yaml
EliSchleifer: done. not a nit - it's critical

PR #903, opened 2024-10-30
Example of how to enable advanced logging for toolbox

EliSchleifer:


.trunk/config/log4rs.yaml
```diff
@@ -0,0 +1,17 @@
+appenders:
+  stdout:
+    encoder:
+      pattern: "{d(%H:%M:%S)} | {({l}):5.5} | {f}:{L} | {m}{n}"
+    kind: console
+
+  file:
+    kind: file
+    path: { REPLACE_THIS_WITH_PATH_WHERE_YOU_WANT_LOGFILE }
```
EliSchleifer, line 9: This needs to be an absolute path to where you want the log to be written to.

PR #914, opened 2024-11-12
Fix ruff parsing for source files with syntax errors

det:
Ruff now reports syntax errors as a diagnostic with a `null` code.


TylerJang27 (approved):
LGTM

TylerJang27 (approved):

PR #919, opened 2024-11-18
Update ruby-build version to latest

amkisko:


amkisko:
As additional question: how to patch runtime definition within project configuration? So let's say we would like to have completely custom ruby-build version and download path.

amkisko:
Final working version:
```
downloads:
  - name: custom-ruby-build
    version: 20241105
    downloads:
      - os:
          linux: linux
          macos: macos
        url: https://github.com/rbenv/ruby-build/archive/refs/tags/v20241105.tar.gz
        strip_components: 1
runtimes:
  enabled:
    - ruby@3.1.4
  definitions:
    - type: ruby
      download: custom-ruby-build
```

PR #920, opened 2024-11-19
(Fix): Use new slack schema

TylerJang27:
[`v2.0.0`](https://github.com/slackapi/slack-github-action/releases/tag/v2.0.0) changed the payload schema to use YAML, along with a few other schema changes.

I also disabled the nightly Windows linter tests since they're not providing any signal atm.

Also rolls back an eslint upgrade from https://github.com/trunk-io/plugins/pull/918

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -75,22 +75,19 @@ jobs:
   # This job is used to update the list of validated versions
   linter_tests_release:
     name: Plugin Tests Release
-    # runs-on: [self-hosted, "${{ matrix.os }}"] TODO(Tyler): Set after Windows self-hosted are established.
     runs-on: ${{ matrix.os }}
     timeout-minutes: 150
     strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly.yaml
```diff
@@ -75,22 +75,19 @@ jobs:
...
     name: Plugin Tests Release
-    # runs-on: [self-hosted, "${{ matrix.os }}"] TODO(Tyler): Set after Windows self-hosted are established.
     runs-on: ${{ matrix.os }}
     timeout-minutes: 150
     strategy:
       fail-fast: false
       matrix:
         linter-version: [Snapshots, Latest]
-        os: [ubuntu-latest, macOS, windows-latest]
+        os: [ubuntu-latest, macOS]
```
TylerJang27, line 84: This has been failing rather noisily. Until we can invest time to burn it down, I'm disabling it

det (approved):

PR #943, opened 2024-12-27
Ignore *.dockerignore files as they do not conform to the Dockerfile spec

NellyWhads:
[Docker supports ignore files](https://docs.docker.com/build/concepts/context/#dockerignore-files) to keep build contexts clean.

The current configuration ignores the generic `.dockerignore` as expected, [but attempts to lint Dockerfile-specific configurations](https://docs.docker.com/build/concepts/context/#filename-and-location) such as `Dockerfile.dockerignore`, `test.Dockerfile.dockerignore` and `Dockerfile.build.dockerignore`.

This PR aims to resolve this bug.


TylerJang27 (approved):
Good catch! Thanks again!

Issue #941, opened 2024-12-27
Terraform + Tofu Plugins Bug: `terraform fmt` + `tofu fmt` erroneously run against `**.tf.json` files


Gowiem:
## Info

When committing / changing our `backend.tf.json` file (which is a standard file in both tofu + terraform), both plugins are trying to run their `fmt` command against those files. This results in the following error: 

```txt
trunk_cli_version: 1.22.8
title: terraform exited with exit_code=2
report:
  - terraform exited with exit_code=2
  - linter:
      command: |
        /<MY_USER_PATH>/.cache/trunk/tools/terraform/1.1.0-6bc9ed0aa441bfb0bf13ee36e04a2bde/terraform fmt -no-color -
      stdin_path: |
        /private/var/folders/7l/fp_bz3gx6yzg24hw08ww9vp00000gn/T/trunk-501/GlML4Y/root-modules/kms/backend.tf.json
      run_from: /private/var/folders/7l/fp_bz3gx6yzg24hw08ww9vp00000gn/T/trunk-501/GlML4Y
      timeout: 10m
      rerun: (cd /private/var/folders/7l/fp_bz3gx6yzg24hw08ww9vp00000gn/T/trunk-501/GlML4Y; cat /private/var/folders/7l/fp_bz3gx6yzg24hw08ww9vp00000gn/T/trunk-501/GlML4Y/root-modules/kms/backend.tf.json | env -i PATH=/<MY_USER_PATH>/.cache/trunk/tools/terraform/1.1.0-6bc9ed0aa441bfb0bf13ee36e04a2bde:/<MY_USER_PATH>/.cache/trunk/tools/terraform/1.1.0-6bc9ed0aa441bfb0bf13ee36e04a2bde HOME=/<MY_USER_PATH> /<MY_USER_PATH>/.cache/trunk/tools/terraform/1.1.0-6bc9ed0aa441bfb0bf13ee36e04a2bde/terraform fmt -no-color -)
      affects_cache:
        []
      direct_configs:
        []
      exit_status: exited
      exit_code: 2
      stdout: (none)
      stderr: |

        Error: Argument or block definition required

          on <stdin> line 1:
           1: {

        An argument or block definition is required here.


    parser: |
      (none)
```

If you run `terraform fmt` against that file directly via `terraform fmt -no-color backend.tf.json` then you get the following: 

> Error: Only .tf, .tfvars, and .tftest.hcl files can be processed with terraform fmt

[...]

PR #945, opened 2024-12-30
(Fix): Remove tf.json filetype from terraform

TylerJang27:
Fixes https://github.com/trunk-io/plugins/issues/941. Verified repro with terraform and tofu format on `x.tf.json` files. Did a bit of research as well to make sure our other security scanners don't scan `.tf.json` files, and we should be fine.


det (approved):
Thanks!

PR #951, opened 2025-01-07
Update Biome configuration file with additional supported languages

andreilgeorgescu:
I've added [all the languages that are currently fully supported by Biome](https://biomejs.dev/internals/language-support/).


det (approved):
Thank you!

PR #953, opened 2025-01-08
Add support to biome for linter safe fixes and import organization as…

andreilgeorgescu:
… apart of the `fmt` command


TylerJang27 (approved):
Thanks for the PR!

I confirmed some details offline, and this is a nice improvement for us!

PR #966, opened 2025-01-27
feat: adds terraform-docs action ⚡

Gowiem:
## Info

* Adds terraform-docs as an action to support automatically generating + updating terraform module documentation for terraform projects.
  * Closes https://github.com/trunk-io/plugins/issues/681
  * This is working well locally.
* Adding terraform-docs support via this PR in favor of #946
* Question: Is there a way to ensure that trunk installs `terraform-docs`? I only see a [`tools` declaration for `commitizen`](https://github.com/trunk-io/plugins/blob/main/actions/commitizen/plugin.yaml#L13-L24) and I'm unsure if I can enforce that the user is using [tools/terraform-docs](https://github.com/trunk-io/plugins/tree/main/tools/terraform-docs). What am I missing there?

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ trunk actions enable {action}
 | [`go-mod-tidy-vendor`](actions/go-mod-tidy-vendor/README.md)                                                                                                                     | automatically tidy and vendor go.mod file                  |
 | [`git-blame-ignore-revs`](actions/git-blame-ignore-revs/README.md)                                                                                                               | automatically configure git to use .git-blame-ignore-revs  |
 | [`npm-check`](actions/npm-check/README.md)                                                                                                                                       | check whether NPM installation is up to date               |
+| [`terraform-docs`](actions/terraform-docs/README.md)                                                                                                                             | generate documentation from Terraform modules              |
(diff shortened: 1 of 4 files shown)


TylerJang27 (approved):
Looks good to me! Thank you for your patience through all this!

Unfortunately we don't yet support referencing tools in Trunk Actions, only package tools. This is a known shortcoming atm. The best you can do is something similar to the commitzen definition, and optionally in your script:
1. Check for the presence of `.trunk/tools/terraform-docs` or `.trunk/tools/terraform-docs.ps1`
2. If present, evaluate it directly
3. If `terraform-docs` is in the PATH, use that
4. Otherwise call `trunk tools install terraform-docs`

I'll leave it up to you if you want to add that to your script, this is good to land with or without that change.
6.

Gowiem:
Gotcha. For now... let's :shipit: and see if others end up using this. I'll ship and then circle around to one of our bigger projects tomorrow to update it to use this action that way my team can weigh in on how it works. We'll circle back if we need to make any fixes. 

Thanks for the help moving this forward @TylerJang27!

TylerJang27:
Sounds great. I'm going to try and get an official release out soon, but in the meantime you can always set the `ref` of the plugin source to use the sha when this lands on main

PR #967, opened 2025-01-29
fix: use python3 as executable for tf-docs script

Gowiem:
## Info

* Follow up from #966 - Executing `python` doesn't do the trick, the executable needs to be `python3`


Gowiem:
@TylerJang27 let's get this in before you cut a new release!

TylerJang27 (approved):
Ah sorry I missed this in my first pass

PR #993, opened 2025-03-10
Update kube-linter to 0.7.2

janisz:


TylerJang27 (approved):
Thanks for the PR! Most of our linters are upgraded automatically when tools release new versions, but this is one of the exceptions. We should have this landed soon 🙏

PR #999, opened 2025-04-02
(Fix): Change dockerfile regex to match re2

TylerJang27:
Reverts #943 and also adds an ignore to handle the original request

Fixes the error log `Error parsing '(?i)(?:^|/)Dockerfile\.(?!.*\.dockerignore$).+$': invalid perl operator: (?!`, since re2 doesn't support negative matches. Tested that the log is no longer there and `trunk check query` produces the correct LQRs


EliSchleifer:
nice

PR #1004, opened 2025-04-08
(Debug): Break tests for testing purposes

TylerJang27:
1. prettier: don't run on yaml files, also false positive of the test erroring on "pretty" linters
2. biome: hard-coded to fail and throw error
3. pylint: no change
4. ruff: snapshot changed


det (approved):

TylerJang27:
Break

PR #1008, opened 2025-04-11
Add variants to jobs

cmillar-trunk:
Adds variants to the staging jobs. Only applies to staging, and uses a branch to avoid early public availability.


TylerJang27 (commented):
You'll probably also want to wire in changing places where `JEST_JUNIT_SUITE_NAME` is set, otherwise the variants will only show up under different tests still

cmillar-trunk:
> You'll probably also want to wire in changing places where `JEST_JUNIT_SUITE_NAME` is set, otherwise the variants will only show up under different tests still

I'm not sure I understand this? Why does that matter, and what do you mean by show up under different tests?

cmillar-trunk:
> So if the desire of this change is for this to show up as one `Testing linter osv-scanner main Snapshots` instead of separate `Testing linter osv-scanner <os> main Snapshots`, you should modify the use of `JEST_JUNIT_SUITE_NAME` to omit the OS piece. This will create a _new_ test to track this, but I think that's fine

@TylerJang27 Ah, gotcha. Since we're only testing this on staging, and the uploader runs for the same tests on both staging and prod, I'd prefer to leave it as-is to not mess with prod for now, if that's alright?

TylerJang27 (approved):
Sounds good, that's fine by me

PR #1027, opened 2025-05-14
(Chore): Fix repo tests

TylerJang27:
Mainly snapshot updates. Bumps node version too


linters/semgrep/semgrep.test.ts
TylerJang27, line 1: The verbose output of semgrep includes this bit:
```txt
          Skipped by --include patterns:

           • test_data/basic.go
           • test_data/element.ts
           • test_data/empty_go.go
           • test_data/empty_js.js
           • test_data/empty_py.py
           • test_data/request.py
```
despite the includes having those exact files. Subsequent runs do not have this problem. I tried getting around this with various flags like `--no-git-ignore --verbose` but to no avail. I don't have a more isolate repro handy since this seems to happen only after startup/with fresh test dirs, but it should probs be reported to semgrep eventually

puzzler7 (approved):

PR #1038, opened 2025-05-28
Update `known_good_version` of mypy

yu-iskw:


EliSchleifer:
thanks. this is landed now

PR #1063, opened 2025-06-19
biome: run from parent to fix silently broken plugin

dex4er:
It addresses problems:

* `run_from: ${parent}`

Now plugin breaks with the error:

```
      direct_configs:
        - biome.jsonc
      exit_status: exited
      exit_code: 1
      stdout: (none)
      stderr: |
        /Users/dex4er/Sources/freelens/freelens/biome.jsonc configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

          × Found a nested root configuration, but there's already a root configuration.
          
          i The other configuration was found in /private/var/folders/k5/wyp40c2n6kzfqm5_gzry576c0000gn/T/trunk-501/MbkdUy/GW6i9I.
          
          i Use the migration command from the root of the project to update the configuration.
          
          $ biome migrate --write
          

        configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

          × Biome exited because the configuration resulted in errors. Please fix them.
```

With this setting biome correctly detects errors.

* `files: html`

HTML file types are supported by Biome from 2.0 when enabled in configuration.


dex4er:
So it must be just `${parent}` otherwise I get

```
          × Found a nested root configuration, but there's already a root configuration.
          
          i The other configuration was found in /private/var/folders/k5/wyp40c2n6kzfqm5_gzry576c0000gn/T/trunk-501/cMNcwD/7iasVC.
          
          i Use the migration command from the root of the project to update the configuration.
          
          $ biome migrate --write
          

        configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

          × Biome exited because the configuration resulted in errors. Please fix them.
```

dex4er:
The problem is related to https://github.com/biomejs/biome/pull/6662 which was not really fixed yet and decribed in https://github.com/biomejs/biome/issues/6509

Biome sees the copy of working tree in sandbox as duplicated then refuses to work.

My workaround is to use explicit `--config-path` option. Unfortunately, we must use explicit config file name, then it is just hardcoded to `biome.jsonc` as I can be just a single file in this option.

I disabled cache as a first reason of the problem that it was for long time undetected. Biome is so fast that caching is not important.

dex4er:
FYI: I had a lot of troubles with using Biome and Trunk together. In the end my configuration now is: https://github.com/freelensapp/freelens/blob/a4d0d192398878caee3c73d9721f60071fdf9121/.trunk/trunk.yaml#L22

```yaml
lint:
  definitions:
    # See https://github.com/trunk-io/plugins/pull/1063
    - name: biome
      files:
        - astro
        - css
        - graphql
        - html
        - javascript
        - json
        - typescript
      commands:
        - output: regex
          success_codes:
            - 0
            - 1
          batch: true
          cache_results: false
          name: lint
          parse_regex: ^::(?P<severity>.*) title=(?P<code>[^,]+),file=(?P<path>[^,]+),line=(?P<line>[^,]+),endLine=[^,]+,col=(?P<col>[^,]+),endColumn=[^:]+::(?P<message>.*)
          read_output_from: stdout
          run: biome check --config-path=${workspace}/biome.jsonc --reporter=github ${target}
          run_from: ${root_or_parent_with(biome.jsonc)}
        - output: regex
          success_codes:
            - 0
          batch: true
          cache_results: false
          formatter: true
          name: fmt
          parse_regex: ^::(?P<severity>.*) title=(?P<code>[^,]+),file=(?P<path>[^,]+),line=(?P<line>[^,]+),endLine=[^,]+,col=(?P<col>[^,]+),endColumn=[^:]+::(?P<message>.*)
          read_output_from: stdout
          run: biome format --config-path=${workspace}/biome.jsonc --reporter=github --write ${target}
          run_from: ${root_or_parent_with(biome.jsonc)}
      run_timeout: 1m
```

so it is now just `${root_or_parent_with(biome.jsonc)}`.

And the main problem is that if biome uses `vcs: {useIgnoreFile: true}` option AND the `.gitignore` file contains `tmp` path then Biome does not work with Trunk Check!

Actually this is the same with `yamlfmt` tool: `tmp` cannot be in `.gitignore`.

EliSchleifer:
@dex4er - yeah go ahead and update the PR. I do think better to pull out the changes by version so it will not affect legacy setups.

PR #1070, opened 2025-07-24
feat: run terraform-docs in directories with terraform file changes

chris3ware:
I have attempted to update the terraform-docs action to support mono repos, by introducing functionality to find directories where terraform files have changed. 

The action runs `terraform-docs` in each directory - using the config file from the root of the repository if it exists, or with default values if it does not.

The native `--recursive` flag does not perform well in this type of scenario and I thought this could be a better solution.


EliSchleifer:
thanks chris. I integrated these ideas and added a test for it.

PR #1074, opened 2025-08-12
Add Jpg, gif, pcap, tif, webp as file extensions

Ryang20718:
Add these following extensions. In our monorepo, we'd like to error on these files being added under git lfs and we'd like to make an upstream PR to remove these file definitions from our local trunk yaml config!


TylerJang27 (approved):
Thanks for the PR! LGTM

I'll get this released in the next couple days

PR #1078, opened 2025-08-21
fix(cspell): add all missing but supported config paths

casaper:
I realised that cspell.config.mjs was not supported in my project.

Considering [cspell's possible config file paths](https://cspell.org/docs/getting-started#1-create-a-configuration-file), I realised that there are actually a lot and most are missing in the plugin.

When these paths are added, the plugin should work again, in any possible project setup.

### DIFF
--- a/linters/cspell/plugin.yaml
+++ b/linters/cspell/plugin.yaml
@@ -77,7 +77,6 @@ lint:
         - cspell.jsonc
         - cspell.yaml
         - cspell.yml
-        - package.json
       affects_cache:
         - package.json
       known_good_version: 7.0.0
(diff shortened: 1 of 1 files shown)


linters/cspell/plugin.yaml
```diff
@@ -24,17 +24,60 @@ lint:
...
+        - cspell.config.jsonc
+        - cspell.config.mjs
+        - cspell.config.toml
         - cspell.config.yaml
         - cspell.config.yml
+        - cspell.json
+        - cspell.jsonc
         - cspell.yaml
         - cspell.yml
+        - package.json
```
TylerJang27, line 80:
```suggestion
        - package.json
```

TylerJang27 (commented):
Thanks for the PR and thanks for catching this! The one change I ask you make is to remove `package.json` from the `direct_configs`. It will still properly be used since it's enumerated in `affects_cache`, but for best practice we set `direct_configs` items to only configs unique to that linter

PR #1117, opened 2026-02-18
fix(uv): validate config with ‎`uv lock --check`

katybaulch:
Change `uv check` to `uv lock --check` to fix the `uv-check` action - as `uv` does not have a `uv check` command  currently (see [here](https://github.com/astral-sh/uv/issues/9653)).


TylerJang27 (approved):
Thanks for the PR!

PR #1134, opened 2026-05-07
On MacOS use shellcheck without rosetta

helly25:
On MacOs use shellcheck on aarch64 without rosetta. There is simply no reason to use rosetta and not everyone has or wants to install it.


EliSchleifer (approved):
LGTM

PR #1135, opened 2026-05-08
biome: fix silent failures (warnings + .gitignore in sandbox)

gejustin:
## Summary

Two independent fixes for biome's lint integration that today cause findings to be silently dropped. Each commit stands alone.

### 1. Capture warning-level findings, anchor diagnostic block

`parse_regex` matched only `×` markers — every `!` warning dropped silently. It also lazily spanned newlines, occasionally correlating one diagnostic's path/line/col with a different diagnostic's message.

The new regex anchors each diagnostic to its own `header → blank line → marker` so captures stay self-contained, and accepts `×` and `!` as terminators. Info-level `i` is intentionally excluded — biome surfaces those as hints, not gating.

`basic_check` snapshot updated to the actual warning biome emits on the fixture (previously captured a cross-correlated result).

New `warning_check` test enables `noNonNullAssertion: warn` in `biome.json` to verify warnings reach the snapshot.

### 2. Symlink .gitignore into the lint sandbox

`vcs.useIgnoreFile: true` in `biome.json` makes biome require the workspace's `.gitignore` at its CWD. trunk's per-target sandbox doesn't include it, so biome aborts with `internalError/fs`, exits 1, and `success_codes: [0, 1]` swallows the error — every finding silently dropped.

Add a `symlinks:` directive that lands `\${workspace}/.gitignore` at the sandbox root next to `biome.json` (the only documented escape hatch — same pattern as remark-lint's `node_modules`). Covers a single root-level `.gitignore`; nested ignore files would each need their own entry (no glob form supported). Harmless when `useIgnoreFile` is unset.

New `gitignore_check` test sets `vcs.useIgnoreFile: true` + writes a `.gitignore`, then expects biome's stock `noUnusedVariables` finding. Adversarially verified: stash `plugin.yaml`, the test fails with `issues: []`.

## Test plan

- [x] All 5 biome tests pass at `KnownGoodVersion` (2.3.4): `basic_check`, `warning_check`, `gitignore_check`, `basic_fmt`, `basic_json`
- [x] Adversarial: stash regex change → `warning_check` fails with empty issues
- [x] Adversarial: stash symlinks change → `gitignore_check` fails with empty issues
- [x] Validated end-to-end against a downstream repo (joinhandshake/joinera) two ways:
  - As a `trunk.yaml` override of the published plugin (catches the violation that previously slipped CI)
  - With `plugins.sources` pointing at this branch (same result)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/linters/biome/biome.test.ts
+++ b/linters/biome/biome.test.ts
@@ -26,3 +26,24 @@ linterCheckTest({
   namedTestPrefixes: ["warning_check"],
   preCheck: enableWarnRule,
 });
+
+// Verify .gitignore reaches the sandbox so vcs.useIgnoreFile works. Without
+// the symlinks: directive in plugin.yaml, biome aborts with internalError/fs
+// when it can't find the ignore file and silently drops all findings.
+// Regression coverage for trunk-io/plugins#1113 issue 2.
+const setupGitignoreRepo: TestCallback = (driver) => {
+  driver.writeFile(
+    "biome.json",
+    JSON.stringify({
+      $schema: "https://biomejs.dev/schemas/2.3.4/schema.json",
+      vcs: { clientKind: "git", enabled: true, useIgnoreFile: true },
+      linter: { enabled: true, rules: { recommended: true } },
+    }),
+  );
+  driver.writeFile(".gitignore", "");
+};
+linterCheckTest({
+  linterName: "biome",
+  namedTestPrefixes: ["gitignore_check"],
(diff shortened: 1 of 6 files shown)


EliSchleifer (approved):
LGTM

EliSchleifer:
@gejustin - this looks good to me. anything else you want to fix here? its marked as draft

PR #1149, opened 2026-07-17
Add linter: toml-tidy

AndrewDongminYoo:
## What

Adds [toml-tidy](https://github.com/AndrewDongminYoo/toml-tidy) as a TOML formatter: it sorts keys within each table (and sibling table declarations) while preserving table hierarchy, comments, and source formatting via tomlkit.

## Why

It complements the existing `taplo` integration (whitespace/style formatting) rather than replacing it — the relationship mirrors `prettier` + `sort-package-json` for `package.json`, which this repo already ships. Sorted TOML keeps files like `pyproject.toml` diff-friendly and merge-conflict-resistant.

## Definition notes

- `runtime: python` / PyPI package, following the `black`/`isort` pattern — no binary downloads to maintain.
- `suggest_if: never` — key sorting is an opinionated choice, so it stays opt-in.
- `affects_cache: [pyproject.toml]` — per-file defaults live in `[tool.toml-tidy]`.
- toml-tidy requires Python >= 3.12 (newer than the default hermetic runtime), so the linter README documents enabling `python@3.12.2`, and the test pins it via `preCheck` — same pattern as ktlint's jdk pin.

## Testing

- `npm test linters/toml-tidy` — passes; fmt snapshot generated for `known_good_version` 0.2.0. The fixture exercises key sorting with comment attachment, natural numeric ordering (`item2` < `item10`), sibling-table sorting, and array-of-tables element-order preservation.
- `npm test tests/repo_tests/readme_inclusion.test.ts` — passes (README table updated).

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ trunk check enable {linter}
 | Terraform       | [terraform] (validate and fmt), [checkov], [tflint], [tfsec], [terrascan], [tofu]                                                        |
 | Terragrunt      | [terragrunt]                                                                                                                             |
 | Textproto       | [txtpbfmt]                                                                                                                               |
-| TOML            | [taplo]                                                                                                                                  |
+| TOML            | [taplo], [toml-tidy]                                                                                                                     |
 | Typescript      | [deno], [eslint], [prettier], [rome], [semgrep]                                                                                          |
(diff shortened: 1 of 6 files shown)


TylerJang27 (approved):
Thanks for the PR!

